PR #3, opened 2025-05-26
(cleanup): Remove unnecessary default files and code from intitial module setup

nsalifu:
This PR removes unnecessary files and code that were included by default during module creation. The cleanup ensures the project only contains relevant source files and configuration, making the codebase cleaner and easier to maintain. No functional changes have been made.


nsalifu:
@wikumChamith, @ManojLL. here is the clean up PR

api/pom.xml
```diff
@@ -11,41 +11,4 @@
 	<packaging>jar</packaging>
 	<name>Auditlogweb API</name>
 	<description>API project for Auditlogweb</description>
-
-	<dependencies>
-		<dependency>
```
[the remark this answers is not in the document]
nsalifu: just pushed a revert

.idea/vcs.xml
```diff
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="VcsDirectoryMappings" defaultProject="true" />
+  <component name="VcsDirectoryMappings">
```
wikumChamith, line 3: Do we need this file?

nsalifu: > Do we need this file?

I think this is was added by the IDE.   Let me remove

PR #11, opened 2025-06-03
(AUDIT-20): Added JSPs for displaying Envers audits

nsalifu:
Issue worked on: https://openmrs.atlassian.net/browse/AUDIT-20

#### PR DESCRIPTION:
This pull request introduces a JSP view for displaying Envers audit logs with improved UI and user experience. Key changes include:

Added controllers and JSPs to fetch and display audit logs.
Added input fields for filtering by user and date (from/to).
Improved layout and spacing for better usability and alignment with the application's design.

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
@@ -95,7 +95,7 @@ public class UtilClass {
             boolean failedCurr = false;
 
             try {
-                currVal = currentEntity != null ? String.valueOf(field.get(currentEntity)) : null;
+                currVal =  String.valueOf(field.get(currentEntity));
             } catch (Exception e) {
                 log.warn("Failed to read current value of field '{}': {}", field.getName(), e.getMessage());
                 failedCurr = true;


wikumChamith:
> Hi @wikumChamith @ManojLL , kindly checkout this too.
> 
> I am not sure if adding both the js and CSS insame jsp file or we separate into different files? I took this inspiration of including all in one file from other modules, like the rest web services and legacyui modules.
> 
> For the UI, do you think maintaining the structure is good ? I did not know a better way for display audit logs other than on tables just as the initial work followed.


I prefer keeping CSS and JS in separate files. Also, I think using Thymeleaf is a better choice than plain JSP. We're already doing that in the FHIR2 module:

[https://github.com/openmrs/openmrs-module-fhir2/blob/master/omod/src/main/resources/org/openmrs/module/fhir2/narratives/AllergyIntolerance.htm](https://github.com/openmrs/openmrs-module-fhir2/blob/master/omod/src/main/resources/org/openmrs/module/fhir2/narratives/AllergyIntolerance.html)

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/ClassUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
+import org.springframework.core.io.support.ResourcePatternResolver;
+import org.springframework.core.type.classreading.CachingMetadataReaderFactory;
+import org.springframework.core.type.classreading.MetadataReader;
+import org.springframework.core.type.classreading.MetadataReaderFactory;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.SystemPropertyUtils;
+import java.io.IOException;
+import java.util.*;
```
wikumChamith, line 13: Wildcard import.

nsalifu: ooh, this file is even not supposed to be here.
The PR for ClassUtil is merged already. updating the PR

wikumChamith:
Can you create a ticket for this in the JIRA and link it here?

wikumChamith:
> > > Hi @wikumChamith @ManojLL , kindly checkout this too.
> > > I am not sure if adding both the js and CSS insame jsp file or we separate into different files? I took this inspiration of including all in one file from other modules, like the rest web services and legacyui modules.
> > > For the UI, do you think maintaining the structure is good ? I did not know a better way for display audit logs other than on tables just as the initial work followed.
> > 
> > 
> > I prefer keeping CSS and JS in separate files. Also, I think using Thymeleaf is a better choice than plain JSP. We're already doing that in the FHIR2 module:
> > [https://github.com/openmrs/openmrs-module-fhir2/blob/master/omod/src/main/resources/org/openmrs/module/fhir2/narratives/AllergyIntolerance.htm](https://github.com/openmrs/openmrs-module-fhir2/blob/master/omod/src/main/resources/org/openmrs/module/fhir2/narratives/AllergyIntolerance.html)
> 
> Yeah! That's right, using Thymeleaf is better. But, how about we continue using JSP for this module, since there is already some groundwork laid out using JSP?

What you’re doing right now is essentially laying the groundwork for this project, so it makes sense to use the most robust and future-proof approach from the start.

omod/pom.xml
```diff
@@ -75,9 +75,6 @@
 			<resource>
 				<directory>src/main/webapp</directory>
 				<filtering>false</filtering>
-				<excludes>
-					<exclude>resources</exclude>
-				</excludes>
```
wikumChamith, line 80: Why was this removed?

nsalifu: > Why was this removed?

Because, I separated the static files, CSS and JS into omod/webapp/resources folder. 
Leaving would always skip loading the resources folder

wikumChamith:
@nsalifu I am getting this error when running these changes.

`Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.openmrs.module.auditlogweb.api.AuditService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}`

nsalifu:
> @nsalifu I am getting this error when running these changes.
> 
> `Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.openmrs.module.auditlogweb.api.AuditService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}`

Yeah that is right, 
please can you add @Service annotation on the AuditServiceImpl class.
I will fix that in a separate PR

wikumChamith:
> > @nsalifu I am getting this error when running these changes.
> > `Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.openmrs.module.auditlogweb.api.AuditService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}`
> 
> Yeah that is right, please can you add `@service` annotation on the AuditServiceImpl class for testing. I will fix that in a separate PR

Where is the PR?

wikumChamith:
image
I am seeing this error.

wikumChamith:
@nsalifu clicking on the view audit trails option on the admin list is giving a 404 error. Can you fix that here?

nsalifu:
> @nsalifu clicking on the view audit trails option on the admin list is giving a 404 error. Can you fix that here?

I think you are pointed to a diff URL.
pointing to this:  /openmrs/module/auditlogweb/auditlogs.form 
should give you `Invalid request parameters` and that's because no entity is selected.

wikumChamith:
image
I am getting this error when I select an audit entity from the list.

[Screencast from 2025-06-12 15-46-23.webm](https://github.com/user-attachments/assets/ae3c4855-2997-4d62-a797-bfdd41a1eac5)

nsalifu:
> image I am getting this error when I select an audit entity from the list.
> 
>  Screencast.from.2025-06-12.15-46-23.webm

@wikumChamith I am not getting this error on my side.

nsalifu:
@wikumChamith this is on side:

https://github.com/user-attachments/assets/c31229ad-c3a1-408d-8f89-24d396933d7f

nsalifu:
> > > @nsalifu clicking on the view audit trails option on the admin list is giving a 404 error. Can you fix that here?
> > 
> > 
> > I think you are pointed to a diff URL. pointing to this: /openmrs/module/auditlogweb/auditlogs.form should give you `Invalid request parameters` and that's because no entity is selected.
> 
> I referring to the option in here
> 
> image

I see the issuw, sorry.
Lemme fix that now.

nsalifu:
> > > > @nsalifu clicking on the view audit trails option on the admin list is giving a 404 error. Can you fix that here?
> > > 
> > > 
> > > I think you are pointed to a diff URL. pointing to this: /openmrs/module/auditlogweb/auditlogs.form should give you `Invalid request parameters` and that's because no entity is selected.
> > 
> > 
> > I referring to the option in here
> > image
> 
> I see the issuw, sorry. Lemme fix that now.

@wikumChamith fixed!

nsalifu:
> > image I am getting this error when I select an audit entity from the list.
> > Screencast.from.2025-06-12.15-46-23.webm
> 
> @wikumChamith I am not getting this error on my side.

I have seen why this caused, can I push the fix here still ?

nsalifu:
> > > image I am getting this error when I select an audit entity from the list.
> > > Screencast.from.2025-06-12.15-46-23.webm
> > 
> > 
> > @wikumChamith I am not getting this error on my side.
> 
> I hadn’t enabled auditing on my side; that was the issue. Would it be possible to add a clearer warning message when auditing isn’t enabled?

It's not clear me to how i can do that, please any pointers ?

nsalifu:
@wikumChamith is it okay, i clean up this PR to just the jsp basic views. we merge it and then on the PR of controllers will handle the exceptions possible ?

nsalifu:
> @nsalifu is this now ready for review?

Yes, marking it as ready for review.

omod/src/main/webapp/viewAudit.jsp
```diff
@@ -0,0 +1,84 @@
...
+                }
+            } catch (IllegalArgumentException ie) {
+                if (ie.getCause() instanceof org.hibernate.QueryException) {
+                    out.println("<tr class='audit-row'><td colspan='3' style='color: #b94a48; background: #f2dede;'>Some referenced data could not be loaded." +
+                            " Please contact your administrator if this problem persists.</td></tr>");
+                } else {
+                    throw ie;
+                }
+            }
+        %>
```
wikumChamith, line 80: Would it be possible to avoid the use of Java in this JSP file? This was a shortcut I took back then, and not a good practice.

wikumChamith (changes requested):

omod/src/main/webapp/viewAudit.jsp
nsalifu: > Would it be possible to avoid the use of Java in this JSP file? This was a shortcut I took back then, and not a good practice.

Okay, I will try to refactor that to the controller. I actually thought of that, but then I applied the rule, "if it works don't touch it" 😄

wikumChamith:
@nsalifu, when handling `org.hibernate.ObjectNotFoundException:`, would it be possible to display the message only on the field we could not access? While showing the rest of the data.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/EnversUtils.java
```diff
@@ -0,0 +1,10 @@
+package org.openmrs.module.auditlogweb.api.utils;
```
wikumChamith, line 1: licenese header

omod/src/main/java/org/openmrs/module/auditlogweb/web/dto/AuditFieldDiff.java
```diff
@@ -0,0 +1,15 @@
+package org.openmrs.module.auditlogweb.web.dto;
```
wikumChamith, line 1: ditto

omod/src/main/webapp/resources/css/auditlogs.css
```diff
@@ -0,0 +1,99 @@
+
```
wikumChamith, line 1: ditto

omod/src/main/webapp/resources/css/viewAudit.css
```diff
@@ -0,0 +1,70 @@
+.audit-container {
```
wikumChamith, line 1: ditto

omod/src/main/webapp/resources/scripts/auditlogs.js
```diff
@@ -0,0 +1,44 @@
+
```
wikumChamith, line 1: ditto

omod/src/main/webapp/viewAudit.jsp
```diff
@@ -0,0 +1,40 @@
+<%@ include file="/WEB-INF/template/include.jsp" %>
```
wikumChamith, line 1: ditto

ManojLL (changes requested):

omod/src/main/webapp/viewAudit.jsp
nsalifu: @wikumChamith I wondered if we have to add license headers to jsp.
I looked at a example [here](https://github.com/openmrs/openmrs-module-legacyui/blob/48ee43be8f40ef892ce5215739ec91eddeccc8c3/omod/src/main/webapp/index.jsp#L1) and it was not added

wikumChamith: Let's add that if we can just to be safe :)

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/AuditlogwebController.java
```diff
@@ -8,40 +8,54 @@
...
+import java.util.List;
 
 /**
  * This class configured as controller using annotation and mapped with the URL of
  * 'module/${rootArtifactid}/${rootArtifactid}Link.form'.
  */
-@Controller("${rootrootArtifactid}.AuditlogwebController")
-@RequestMapping(value = "module/${rootArtifactid}/${rootArtifactid}.form")
+@Controller
+@RequestMapping(value = "module/auditlogweb/auditlogs.form")
```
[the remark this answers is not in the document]
nsalifu: > Let's try to avoid hardcoding the "module/auditlogweb" part here.

can I use constants to supply it ?
like this:
public class AuditlogwebConstants {
    public static final String MODULE_PATH = "/module/auditlogweb";
    public static final String AUDITLOGS_FORM = MODULE_PATH + "/auditlogs.form";
}
and 

@RequestMapping(value = AUDITLOGS_FORM)
@RequiredArgsConstructor
public class AuditlogwebController {
    private final String VIEW = MODULE_PATH + "/auditlogs";
    private final String ENVERS_DISABLED_VIEW = MODULE_PATH + "/enversDisabled";


or any nicer means to do ? @wikumChamith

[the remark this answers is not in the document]
wikumChamith: Just having the MODULE_PATH = "/module/auditlogweb/" is enough. You could update the request mappings like
MODULE_PATH + "auditlogsweb"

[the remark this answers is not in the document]
nsalifu: > Just having the `MODULE_PATH = "/module/auditlogweb/`" is enough. You could update the request mappings like `MODULE_PATH + "auditlogsweb"`

Addressed here: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/40c059adb74dd37edae4dff6f4c7d06a2c34cbef

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,144 @@
...
+        try {
+            return obj.toString();
+        } catch (org.hibernate.ObjectNotFoundException | org.hibernate.LazyInitializationException e) {
+            return "Data not found";
+        } catch (Exception e) {
+            return "Data not found";
+        }
+    }
+
+}
```
wikumChamith, line 144: Could we move this to a different location? This does not feel like it should belong in a controller class.

wikumChamith (changes requested):

wikumChamith (changes requested):

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.List;
+import java.lang.reflect.Field;
+import java.util.Objects;
+
+@Controller
+@RequestMapping(value = ViewAuditController.MODULE_PATH + "/viewAudit.form")
+@RequiredArgsConstructor
+public class ViewAuditController {
+
+    public static final String MODULE_PATH = "/module/auditlogweb";
```
wikumChamith, line 32: I thought you were moving this into a separate class? 

https://github.com/openmrs/openmrs-module-auditlogweb/pull/11#discussion_r2149750181

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,144 @@
...
+
+            boolean isDifferent = !Objects.equals(oldStr, currStr);
+
+            diffs.add(new AuditFieldDiff(field.getName(), oldStr, currStr, isDifferent));
+        }
+
+        return diffs;
+    }
+
+    private String safeToString(Object obj) {
```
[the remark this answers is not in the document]
nsalifu: > Why is this method called safeToString??



> Why is this method called safeToString??

Called it this way for just clarity, initially I was calling obj.toString() on comparing the retrieved fields, I noticed at some point it was crashing because some fields are `null`.  Based on what Ian mentioned, this method might not be useful here again. Let me look into it

omod/src/main/webapp/enversDisabled.jsp
```diff
@@ -0,0 +1,17 @@
...
+  graphic logo is a trademark of OpenMRS Inc.
+--%>
+<%@ include file="/WEB-INF/template/include.jsp"%>
+<%@ include file="/WEB-INF/template/header.jsp"%>
+
+<div class="error-container">
+    <h2>Hibernate Envers Not Enabled on the Server</h2>
+    <p>${errorMessage}</p>
+</div>
+
```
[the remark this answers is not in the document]
nsalifu: > It would be better to title this "Audit Logging Not Enabled on this Server". Also, it might be helpful to check if the user is a system administrator and if so, provide them with instructions to enable audit logging.

I have this
Screenshot 2025-06-17 at 13 07 44

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/AuditlogwebController.java
```diff
@@ -8,40 +8,55 @@
...
+import java.util.List;
 
 /**
  * This class configured as controller using annotation and mapped with the URL of
  * 'module/${rootArtifactid}/${rootArtifactid}Link.form'.
  */
-@Controller("${rootrootArtifactid}.AuditlogwebController")
-@RequestMapping(value = "module/${rootArtifactid}/${rootArtifactid}.form")
+@Controller
+@RequestMapping(value = AuditlogwebController.MODULE_PATH + "/auditlogs.form")
```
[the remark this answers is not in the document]
nsalifu: > They don't need to be available at runtime. They should be resolved at build time [(see here)](https://github.com/openmrs/openmrs-contrib-maven-plugin/blob/4ce13e1d8f42d7b293c3db66d46fd36c1a28c678/src/main/java/org/motech/openmrs/plugin/InitializeModuleMojo.java#L162). This is pretty much equivalent to hard-coding it as you've done, but more flexible if we decide we need to rename the module for w/e reason.

okay. Sorry, @ibacher. Now I'm not sure what to do next.  Keep the use of AuditlogwebConstants or try out from what you mentioned above?

omod/src/main/java/org/openmrs/module/auditlogweb/AuditlogwebConstants.java
```diff
@@ -0,0 +1,6 @@
+package org.openmrs.module.auditlogweb;
+
+public final class AuditlogwebConstants {
+    public static final String MODULE_PATH = "/module/auditlogweb";
+    private AuditlogwebConstants() {}
```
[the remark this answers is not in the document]
wikumChamith: No let's keep this :)

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,92 @@
...
+            } catch (org.hibernate.ObjectNotFoundException ex) {
+                model.addAttribute("errorMessage", "Audit entity not found for this revision.");
+                return new ModelAndView(VIEW, model);
+            }
+
+            if (auditId - 1 > 0) {
+                try {
+                    oldEntity = auditService.getRevisionById(clazz, entityId, auditId - 1);
+                } catch (org.hibernate.ObjectNotFoundException ex) {
+                    oldEntity = null;
```
wikumChamith, line 72: Do you think this assignment is necessary?

pom.xml
```diff
@@ -115,6 +115,12 @@
 				<version>${lombokVersion}</version>
 				<scope>provided</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.slf4j</groupId>
+				<artifactId>slf4j-api</artifactId>
+				<version>2.0.17</version>
```
wikumChamith, line 121: Use a property for the version.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -25,20 +30,46 @@ public static List<String> findClassesWithAnnotation() {
...
+
+    public static List<AuditFieldDiff> computeFieldDiffs(Class<?> clazz, Object oldEntity, Object currentEntity) {
+        List<AuditFieldDiff> diffs = new ArrayList<>();
+
+        if (currentEntity == null) {
+            return diffs;
+        }
+        Field[] fields = clazz.getDeclaredFields();
+        for (Field field : fields) {
+            if (java.lang.reflect.Modifier.isStatic(field.getModifiers()) || field.isSynthetic()) {
```
wikumChamith, line 52:
```suggestion
            if (Modifier.isStatic(field.getModifiers()) || field.isSynthetic()) {
```

wikumChamith:
@nsalifu, let's display the edit type on the Audit Log Details page.

wikumChamith:
@nsalifu I noticed a couple of UI issues:

- "Audit Table for" seems incomplete; shouldn’t it display the name of the selected entity (e.g., PersonName)?
- Let's replace ADD and MOD in the Revision Type column with more descriptive labels like CREATION and MODIFICATION.

nsalifu:
> * Let's replace ADD and MOD in the Revision Type column with more descriptive labels like CREATION and MODIFICATION



> @nsalifu I noticed a couple of UI issues:
> 
> * "Audit Table for" seems incomplete; shouldn’t it display the name of the selected entity (e.g., PersonName)?
> * Let's replace ADD and MOD in the Revision Type column with more descriptive labels like CREATION and MODIFICATION.
> 
> image

@wikumChamith Is it better as such now? 
Screenshot 2025-06-18 at 15 06 06

Screenshot 2025-06-18 at 15 06 27

ManojLL (changes requested):

wikumChamith:
@nsalifu add Javadoc comments to methods.

wikumChamith:
@nsalifu I think the audit list should show the most recent entries first by default. We also need to add pagination when retrieving the audit records. Could you give that a try?

nsalifu:
> @nsalifu I think the audit list should show the most recent entries first by default. We also need to add pagination when retrieving the audit records. Could you give that a try?

@wikumChamith Alright, I will give it a try.  I  thought of the pagination but ignored it at this point, thinking that could be more useful when implementing the filtering.

nsalifu:
> @nsalifu I think the audit list should show the most recent entries first by default. We also need to add pagination when retrieving the audit records. Could you give that a try?

@wikumChamith I update the auditlogs to be sorted by most recent(time) and pagination here: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/1e37e486662ce3314bb891ee3ae69522a2c9b54b.
Kindly have a look

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -44,9 +64,60 @@ public <T> List<AuditEntity<T>> getAllRevisions(Class<T> entityClass) {
...
+    public int countAllRevisions(Class<?> entityClass) {
+        AuditReader auditReader = AuditReaderFactory.get(sessionFactory.getCurrentSession());
+
+        AuditQuery countQuery = auditReader.createQuery()
+                .forRevisionsOfEntity(entityClass, false, true)
+                .addProjection(org.hibernate.envers.query.AuditEntity.revisionNumber().count());
+
+        Number countResult = (Number) countQuery.getSingleResult();
+        return countResult != null ? countResult.intValue() : 0;
+    }
```
wikumChamith, line 82: How about using this?

```
    public long countAllRevisions(Class<?> entityClass) {
        AuditReader auditReader = AuditReaderFactory.get(sessionFactory.getCurrentSession());

       return (long) auditReader.createQuery()
                .forRevisionsOfEntity(entityClass, false, true)
                .addProjection(org.hibernate.envers.query.AuditEntity.revisionNumber().count())
                .getSingleResult();
    }
```

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -30,10 +35,25 @@ public AuditDao(SessionFactory sessionFactory) {
         this.sessionFactory = sessionFactory;
     }
```
wikumChamith, line 37: Can't we use Lombok in this class?

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
nsalifu: > How about using this?
> 
> ```
>     public long countAllRevisions(Class<?> entityClass) {
>         AuditReader auditReader = AuditReaderFactory.get(sessionFactory.getCurrentSession());
> 
>        return (long) auditReader.createQuery()
>                 .forRevisionsOfEntity(entityClass, false, true)
>                 .addProjection(org.hibernate.envers.query.AuditEntity.revisionNumber().count())
>                 .getSingleResult();
>     }
> ```

I think this makes more sense,  just checked the hibernate is already returning long something I missed initially. thanks @wikumChamith for the catch.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -21,32 +22,96 @@
...
+                .setMaxResults(size);
+
         return (List<AuditEntity<T>>) auditQuery.getResultList().stream().map(result -> {
             Object[] array = (Object[]) result;
             T entity = entityClass.cast(array[0]);
             OpenmrsRevisionEntity revisionEntity = (OpenmrsRevisionEntity) array[1];
             RevisionType revisionType = (RevisionType) array[2];
-            String changedBy = Context.getUserService().getUser(revisionEntity.getChangedBy()).toString();
-            return new AuditEntity<>(entity, revisionEntity, revisionType, changedBy);
+            return new AuditEntity<>(entity, revisionEntity, revisionType, null);
```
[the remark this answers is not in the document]
nsalifu: > Why is `changedBy` null?

I am loading it at the service level.  Here the constructor expects 4 args that's why I left it to be null, but then later on populated in the service-level call.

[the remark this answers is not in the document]
wikumChamith: I don't think the DAO layer should return it with a null value and rely on the caller to set the changedBy field.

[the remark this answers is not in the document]
nsalifu: > I don't think the DAO layer should return it with a null value and rely on the caller to set the changedBy field.

okay.  any idea  what I should do please  ?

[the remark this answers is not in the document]
wikumChamith: Can't we do a database call here?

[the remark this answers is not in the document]
wikumChamith: Can't we write a Hibernate query to get the value for that field?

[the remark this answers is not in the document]
nsalifu: > What if we update the entity to store the user id rather than storing the username?

I think that will work too, and maintains the Dao just for data retrieval. This means we can use almost the same way as the service-level is doing to make sure we return the actual user from the userId coming from the Dao to the view. 
Let me know if this is the actual flow I should go with. Then I will be looking into that.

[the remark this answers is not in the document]
nsalifu: @wikumChamith here is the approach i followed for this again: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/4879874825305780c541be8281167bb0fbb0cbd1

api/src/test/java/org/openmrs/module/auditlogweb/api/dao/AuditDaoTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.openmrs.api.db.hibernate.envers.OpenmrsRevisionEntity;
+import org.openmrs.module.auditlogweb.AuditEntity;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class AuditDaoTest {
```
wikumChamith, line 34: Is there a reason why you are not using BaseModuleContextSensitiveTest?

nsalifu: @wikumChamith 
I think we don't need to use `BaseModuleContextSensitiveTest` because the intention here is pure unit tests.
I noticed that `BaseModuleContextSensitiveTest` requires a real Spring context/db, making it more integration-like.

api/pom.xml
```diff
@@ -55,6 +53,24 @@
...
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-api</artifactId>
+			<version>5.13.1</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<version>5.13.1</version>
+			<scope>test</scope>
```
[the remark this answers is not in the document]
nsalifu: > Why don't you take a look at the errors again and try to find out what dependency you should import?

thanks @wikumChamith updated to this: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/5102041475f75bccbecc1303f13aa0cf2d03f13f

api/pom.xml
```diff
@@ -12,41 +13,63 @@
...
+					<groupId>org.mockito</groupId>
+					<artifactId>mockito-junit-jupiter</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.powermock</groupId>
+					<artifactId>powermock-module-junit4</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.powermock</groupId>
+					<artifactId>powermock-api-mockito2</artifactId>
```
wikumChamith, line 69: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/files#r2175413969

api/pom.xml
```diff
@@ -12,41 +13,63 @@
...
 			<groupId>org.openmrs.web</groupId>
 			<artifactId>openmrs-web</artifactId>
 			<type>test-jar</type>
 			<scope>test</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>org.mockito</groupId>
+					<artifactId>mockito-core</artifactId>
+				</exclusion>
+			</exclusions>
```
wikumChamith, line 46: We don't need this.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -20,30 +22,130 @@
...
+     */
+    @Override
+    public String resolveUsername(Integer userId) {
+        if (userId == null) {
+            return "Unknown";
+        }
+
+        User user = Context.getUserService().getUser(userId);
+        String username = user.getUsername();
+        if (username == null || username.trim().isEmpty()) {
```
wikumChamith, line 145: Can't we use StringUtils here?

pom.xml
```diff
@@ -115,6 +118,23 @@
...
+				<groupId>org.mockito</groupId>
+				<artifactId>mockito-inline</artifactId>
+				<version>${mockitoInlineVersion}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
+				<groupId>net.bytebuddy</groupId>
+				<artifactId>byte-buddy</artifactId>
+				<version>${byteBuddyVersion}</version>
+			</dependency>
```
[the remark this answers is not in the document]
wikumChamith: Can we add a scope to this?

api/src/test/java/org/openmrs/module/auditlogweb/api/dao/AuditDaoTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.mockito.Mock;
+import org.mockito.MockedStatic;
+import org.mockito.MockitoAnnotations;
+import org.openmrs.api.db.hibernate.envers.OpenmrsRevisionEntity;
+import org.openmrs.module.auditlogweb.AuditEntity;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
```
ManojLL, line 31: wild card imports

api/src/test/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.openmrs.api.UserService;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.auditlogweb.AuditEntity;
+import org.openmrs.module.auditlogweb.api.dao.AuditDao;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
```
ManojLL, line 27: remove this wild card imports

ManojLL (changes requested):

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/EnversUtils.java
```diff
@@ -0,0 +1,55 @@
...
+    public static String getAdminHint() {
+        if (isCurrentUserSystemAdmin()) {
+            return "As a System Administrator, you can enable audit logging by adding the following lines to your"
+                    + " <code>openmrs-runtime.properties</code> file:"
+                    + "<br/><code>hibernate.integration.envers.enabled=true</code>"
+                    + "<br/><code>hibernate.hbm2ddl.auto=update</code>";
+        }
+        return "";
+    }
+}
```
wikumChamith, line 55: Feel like this doesn't belong in this class.

nsalifu: > Feel like this doesn't belong in this class.

That sounds like  a violation of SRP, thanks for noting this out  @wikumChamith. I have refactored to this: https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/797f9d838f45b113a29caf9fc0f22f1b623005d6
Kindly take a look if it's more acceptable.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,136 @@
...
+            String auditType;
+            if (oldEntity == null && currentEntity != null) {
+                auditType = "Record was added";
+            } else if (oldEntity != null && currentEntity == null) {
+                auditType = "Record was deleted";
+            } else if (oldEntity != null && currentEntity != null) {
+                auditType = "Record was modified";
+            } else {
+                auditType = "No change";
+            }
```
wikumChamith, line 110: Can't we use auditEntity.revisionType here?

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
```diff
@@ -0,0 +1,136 @@
...
+            // Compute field differences
+            List<AuditFieldDiff> diffs = UtilClass.computeFieldDiffs(clazz, oldEntity, currentEntity);
+
+            // Determine edit or revision type
+            String auditType;
+            if (oldEntity == null && currentEntity != null) {
+                auditType = "Record was added";
+            } else if (oldEntity != null && currentEntity == null) {
+                auditType = "Record was deleted";
+            } else if (oldEntity != null && currentEntity != null) {
```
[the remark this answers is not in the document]
nsalifu: > Do we need the currentEntity != null here?

@wikumChamith  I think. Technically, if `oldEntity == null`, it implies this is the first revision and the record was added, so checking `currentEntity != null` may be redundant. However, I left it in to be more defensive, in case the audit record is incomplete or corrupted. If we’re confident in the integrity of revisions, I can simplify the logic accordingly.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
nsalifu: > Can't we use auditEntity.revisionType here?

You're right. using `auditEntity.getRevisionType()` would be more reliable and cleaner than inferring the audit type by checking entity nullity. I’ll refactor the code to use RevisionType instead.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -36,9 +55,67 @@ public static List<String> findClassesWithAnnotation() {
...
+
+            field.setAccessible(true);
+
+            String oldVal = null;
+            String currVal = null;
+            boolean failedOld = false;
+            boolean failedCurr = false;
+
+            try {
+                currVal = currentEntity != null ? String.valueOf(field.get(currentEntity)) : null;
```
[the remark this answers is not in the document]
nsalifu: > Do we need currentEntity != null here?

I think removing the `currentEntity != null `check might lead to a NullPointerException when accessing fields on a null object. While the exception is caught, it's better to prevent the exception in the first place with an explicit null check.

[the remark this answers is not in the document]
nsalifu: > I am just saying that we don't need `currentEntity != null` here because we have already filtered out null values in `if (currentEntity == null) return diffs;`. Basically, `currentEntity != null` is always true in here.

Yes,  and also thinking of it deeply, I see that.  For an entity that is deleted, the currentEntity is null, and then the return diffs will skip computing the difference, making not able to actually see the deleted values.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ViewAuditController.java
nsalifu: > Can't we use auditEntity.revisionType here?

For this, I think of doing it this way:  https://github.com/openmrs/openmrs-module-auditlogweb/pull/11/commits/deefc9ebc19b7865d9aa4fad1649a12504a6a3b8

wikumChamith (approved):

PR #12, opened 2025-06-05
(feat): Using thymeleaf for the View layer

nsalifu:
This PR is about migrating from JSP to Thymeleaf.


nsalifu:
Continue using JSP for consistency.

PR #31, opened 2026-03-20
AUDIT-29: Fixes and improvements in displaying of audit logs

druchniewicz:
## Summary

Pull request provides some fixes and improvements in displaying of audit logs.  Please see below screenshot section with short explanation to make it clearer.

## Screenshots
Voided fields are now visible for the user in audit logs table:
image

There were some missing data in some of entities, e.g. you changed the gender for the patient and this change was not visible for the user in audit log table. Now it has been fixed. Example:
image

Instead of multiple null texts what is not very readable for the user we display empty values, example:
image

Inconvenient way of searching for change history. Previously when we e.g. updated encounter (and related obses) and we wanted to verify these changes in audit log table we had to go to the Encounter entity history first, then go to the Obs entity history and search for related Obs changes etc. This is very inconvenient navigation for the user. I added section with related entities changed in given revision. Example:
image

We explore Encounter entity and we can show that two other Obs changes were made in the same revision. We can click “View Audit” to see these changes quickly instead of going back to the main audit log page, selecting Obs entity and searching obses from this revision.

## Related Issues
https://openmrs.atlassian.net/browse/AUDIT-28
https://openmrs.atlassian.net/browse/AUDIT-29
https://openmrs.atlassian.net/browse/AUDIT-30
https://openmrs.atlassian.net/browse/AUDIT-31
https://openmrs.atlassian.net/browse/AUDIT-32

## Other
* Regarding ticket [AUDIT-28](https://openmrs.atlassian.net/browse/AUDIT-28) there also needed changes from this [PR](https://github.com/openmrs/openmrs-core/pull/5933) to backfill audit tables
* Appointments entities are currently not supported by Hibernate envers auditing so we can't track appointments at this moment. I created [ticket](https://openmrs.atlassian.net/browse/AM-241) to adopt appointments module to it.

[AUDIT-28]: https://openmrs.atlassian.net/browse/AUDIT-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


druchniewicz:
@wikumChamith @dkayiwa can you take a look at PR?
I showed examples how it looks like now in related ticket comments

tanyarai5446 (commented):
The implementation of getEntitiesModifiedInRevision using AuditReader is much more efficient than the previous approach. It correctly handles the retrieval of voided entities by querying the revision history directly. This perfectly solves the visibility issue in AUDIT-29.

wikumChamith:
@druchniewicz thanks for the PR. Can you update the PR description with a summary of what you're doing here?

druchniewicz:
> @druchniewicz thanks for the PR. Can you update the PR description with a summary of what you're doing here?

@wikumChamith I updated PR description

druchniewicz:
@wikumChamith @dkayiwa can you take a look at PR?

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -80,6 +90,29 @@ public static boolean doesClassContainsAuditedAnnotation(Class<?> clazz) {
...
+        
+        Class<?> current = clazz;
+        while (current != null && current != Object.class) {
+            Field[] declaredFields = current.getDeclaredFields();
+            for (Field field : declaredFields) {
+                fieldMap.put(field.getName(), field);
+            }
+            current = current.getSuperclass();
+        }
+
```
Binayak490-cyber, line 112: I think there's a bug in the field collision handling. The loop iterates child → parent (current starts at clazz, walks up via getSuperclass()), and calls fieldMap.put(field.getName(), field) at each level. Since put() overwrites existing keys, parent fields encountered later in the loop will overwrite child fields when names collide — which is the opposite of the intended behavior stated in the comment. To fix this, either:

1. Iterate parent → child (reverse order), so child's put() wins, or
2. Use fieldMap.putIfAbsent(field.getName(), field) while keeping the current child-first iteration.

omod/src/main/webapp/viewAudit.jsp
```diff
@@ -29,6 +29,73 @@
...
+                                String className = entityObj != null ? entityObj.getClass().getName() : "";
+                                String simpleName = entityObj != null ? entityObj.getClass().getSimpleName() : "";
+                                Object entityId = null;
+                                try {
+                                    java.lang.reflect.Method method = entityObj.getClass().getMethod("getId");
+                                    entityId = method.invoke(entityObj);
+                                } catch (Exception e) {
+                                    // ignore
+                                }
+                                pageContext.setAttribute("entityIdValue", entityId);
```
Binayak490-cyber: I think using raw <% %> Java scriptlets is discouraged in this codebase, and this one has a concrete safety issue: entityObj.getClass().getMethod("getId") will throw a NoSuchMethodException for any audited entity that doesn't have a getId() method (e.g., GlobalProperty, Role, Privilege). The catch (Exception e) { // ignore } block silently hides this, resulting in a null entityId and a broken "View Audit" link. The existing UtilClass.getEntityIdAsString() already handles this - it should be called in the controller and passed to the view as part of the model, avoiding scriptlet reflection entirely.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -452,4 +452,48 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
...
+     * @param revisionId the revision number to query
+     * @return list of AuditEntity objects modified in this revision
+     */
+    public List<AuditEntity<?>> getEntitiesModifiedInRevision(int revisionId) {
+        List<AuditEntity<?>> result = new ArrayList<>();
+        List<Class<?>> classes = getNonAbstractAuditedClasses();
+
+        AuditReader auditReader = AuditReaderFactory.get(sessionFactory.getCurrentSession());
+
+        for (Class<?> clazz : classes) {
```
Binayak490-cyber: This loop fires one separate AuditQuery per audited entity class. On an OpenMRS instance with many audited types, this could mean 30+ DB queries on every single page load of the audit detail view. Considering batching or caching the result of getEntitiesModifiedInRevision(), or finding a way to query across all entity types in a single call rather than iterating per class.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -452,4 +452,48 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
...
+            try {
+                AuditQuery query = auditReader.createQuery()
+                        .forRevisionsOfEntity(clazz, false, true)
+                        .add(org.hibernate.envers.query.AuditEntity.revisionNumber().eq(revisionId));
+                
+                query.setMaxResults(1000);
```
Binayak490-cyber: setMaxResults(1000) silently truncates results with no warning to the user. If a revision touches more than 1000 rows of a given entity type, the Related Entities panel will silently show incomplete data. Either i think making this limit configurable via a global property, or display a warning in the UI when the result count hits the cap.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -452,4 +452,48 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
...
+                        
+                        if (entity != null) {
+                            Integer userId = revisionEntity != null ? revisionEntity.getChangedBy() : null;
+                            result.add(new AuditEntity<>(entity, revisionEntity, revisionType, userId));
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                log.debug("Could not find revision {} for entity {}: {}", revisionId, clazz.getSimpleName(), e.getMessage());
+            }
```
Binayak490-cyber, line 502: In this i think exceptions are caught and logged only at debug level. In production, operators typically don't enable debug logging, so if an entity class fails to query (e.g., due to a schema mismatch or missing audit table), it will fail completely silently. This should be at least log.warn() so it shows up in standard logs.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -321,6 +322,30 @@ public List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page
...
+        
+        return allEntitiesInRevision.stream()
+                .filter(auditEntity -> {
+                    if (auditEntity.getEntity() == null) {
+                        return false;
+                    }
+
+                    Class<?> actualType = auditEntity.getEntity().getClass();
+                    for (Class<?> fieldType : fieldTypes.values()) {
+                        if (fieldType != null && fieldType.isAssignableFrom(actualType)) {
```
Binayak490-cyber: The isAssignableFrom(actualType) check includes any entity in the revision whose type is assignable from any field type of the main entity. This is too broad i think for example, if Patient has a Person field, and a completely unrelated Person was modified in the same revision, it will appear in this patient's Related Entities panel. The filter can use actual foreign key values or entity references, not just type compatibility, to avoid false positives.

api/src/test/java/org/openmrs/module/auditlogweb/api/utils/UtilClassUnitTest.java
```diff
@@ -65,9 +66,116 @@ public void toEndDate_shouldReturnDateAtEndOfDayOrNull() {
...
+        private String gender;
+        private Date birthdate;
+
+        public Integer getPersonId() { return personId; }
+        public void setPersonId(Integer personId) { this.personId = personId; }
+        public String getGender() { return gender; }
+        public void setGender(String gender) { this.gender = gender; }
+        public Date getBirthdate() { return birthdate; }
+        public void setBirthdate(Date birthdate) { this.birthdate = birthdate; }
+    }
```
Binayak490-cyber: I think PersonImpl is a local stub with only 2 levels of hierarchy and no Hibernate/OpenMRS specifics. The actual bug this test is meant to cover (gender not showing in diffs) was caused by OpenMRS's real inheritance chain: Patient → Person → BaseOpenmrsData → BaseOpenmrsObject. A test using a 2-level stub does not fully guard against that regression. The test can either be use real OpenMRS domain classes or at minimum a 3+ level hierarchy that mirrors the actual structure.

Binayak490-cyber:
Maybe I am wrong @wikumChamith, @druchniewicz, @ManojLL, but can you also check that I am saying it correctly or it is not at all required...!

api/src/test/java/org/openmrs/module/auditlogweb/api/utils/UtilClassUnitTest.java
Binayak490-cyber: As I also faced this.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -452,4 +452,48 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
...
+        }
```
wikumChamith, line 503: Can't we use [AuditReader.getCrossTypeRevisionChangesReader()](https://docs.hibernate.org/orm/7.1/javadocs/org/hibernate/envers/CrossTypeRevisionChangesReader.html) here?

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
```diff
@@ -212,4 +212,16 @@ List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page, int s
...
+     * Finds all entities modified in the same revision as the given entity.
+     * This method retrieves all entities changed in the same revision and filters them
+     * to only include those whose types match the field types of the main entity.
+     *
+     * @param entityClass the main entity class
+     * @param entityId the entity ID
+     * @param revisionId the revision number
+     * @return list of related AuditEntity objects modified in the same revision
+     */
+    List<AuditEntity<?>> getRelatedEntitiesInRevision(Class<?> entityClass, Object entityId, int revisionId);
```
wikumChamith, line 235: Don't we need `@Authorized` here? I think there are some other methods we missed setting the privileges on too. Feel free to add those in this PR.

api/src/test/java/org/openmrs/module/auditlogweb/api/utils/UtilClassUnitTest.java
```diff
@@ -9,17 +9,18 @@
...
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertNotNull;
```
wikumChamith: Is there a reason for switching from Junit5 to Junit4?

omod/src/main/webapp/resources/css/auditlogs.css
```diff
@@ -150,4 +150,10 @@
     color: black;
     font-weight: bold;
     border: 1px solid #ffffff;
+}
+
+.pagination-ellipsis {
+    padding: 8px 12px;
+    color: #666;
+    font-weight: bold;
 }
```
ManojLL:
```suggestion
}

```

omod/src/main/webapp/resources/css/viewAudit.css
```diff
@@ -99,4 +99,81 @@ table.audit-table {
...
+    font-weight: 500;
+    padding: 4px 8px;
+    border-radius: 4px;
+    transition: background 0.15s;
+}
+
+.related-entities-table td a:hover {
+    background: #e8f4fc;
+    text-decoration: underline;
 }
```
ManojLL:
```suggestion
}

```

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
druchniewicz: corrected

omod/src/main/webapp/viewAudit.jsp
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
druchniewicz: removed this max results

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
druchniewicz: it shouldn't be a problem, there should not be revisions with unrelated entities. you will usually change something in scope of one patient, one encounter etc.

api/src/test/java/org/openmrs/module/auditlogweb/api/utils/UtilClassUnitTest.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
druchniewicz: I tried it but I got an error that I need to set org.hibernate.envers.track_entities_changed_in_revision so I set it up but after that I'm getting a lot of issues during app startup. Additionally AI explained that in the current Hibernate version used by OpenMRS (5.6) this API just returns entity objects, so there are missing metadata, revision type or user info. One more thing is that these changes are loaded instantly, there is no performance problem here

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
druchniewicz: corrected

api/src/test/java/org/openmrs/module/auditlogweb/api/utils/UtilClassUnitTest.java
druchniewicz: no, my bad. restored

omod/src/main/webapp/resources/css/auditlogs.css
druchniewicz: fixed

omod/src/main/webapp/resources/css/viewAudit.css
druchniewicz: fixed

druchniewicz:
@wikumChamith I pushed improvements. Could you take a look once again?

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -94,34 +127,38 @@ public static List<AuditFieldDiff> computeFieldDiffs(Class<?> clazz, Object oldE
...
 
-        Field[] fields = clazz.getDeclaredFields();
+        Field[] fields = getAllFields(clazz);
         for (Field field : fields) {
             if (Modifier.isStatic(field.getModifiers()) || field.isSynthetic()) continue;
 
             field.setAccessible(true);
 
-            String oldVal = null;
-            String currVal = null;
```
Binayak490-cyber, line 104: I think the initialisation was correct and not needed anything to change, why it was changed @druchniewicz...?

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
wikumChamith: I think we can optimise this by getting only the relevant fields that can change in the revision.

In the service class we do:

```
Map<String, Class<?>> fieldTypes = UtilClass.getFieldTypes(entityClass);
Set<Class<?>> relevantClasses = new HashSet<>(fieldTypes.values());
List<AuditEntity<?>> allEntitiesInRevision = auditDao.getEntitiesModifiedInRevision(revisionId, relevantClasses);
```

And we can update the DAO like:

```
List<Class<?>> classesToQuery = auditedClasses.stream()
        .filter(c -> relevantClasses.stream().anyMatch(r -> r.isAssignableFrom(c)))
        .collect(Collectors.toList());
AuditReader auditReader = AuditReaderFactory.get(sessionFactory.getCurrentSession());
for (Class<?> clazz : classesToQuery) {
```

This way, if you're viewing an Encounter that has Obs, Patient, and Visit fields, it queries 3 tables instead of all audited entity tables.

omod/src/main/webapp/viewAudit.jsp
```diff
@@ -29,6 +29,67 @@
...
+                    <tbody>
+                        <c:forEach var="related" items="${relatedEntities}">
+                            <%
+                                Object entity = (Object) pageContext.getAttribute("related");
+                                Object entityObj = entity != null ? ((org.openmrs.module.auditlogweb.AuditEntity<?>) entity).getEntity() : null;
+                                String className = entityObj != null ? entityObj.getClass().getName() : "";
+                                String simpleName = entityObj != null ? entityObj.getClass().getSimpleName() : "";
+                                String entityIdValue = entityObj != null ? org.openmrs.module.auditlogweb.api.utils.UtilClass.getEntityIdAsString(entityObj) : "unknown";
+                                pageContext.setAttribute("entityIdValue", entityIdValue);
+                            %>
```
wikumChamith: If we could avoid scriptlet code in JSP, that would be great.

omod/src/main/webapp/viewAudit.jsp
```diff
@@ -29,6 +29,67 @@
...
+                            <tr>
+                                <td><%= simpleName %></td>
```
wikumChamith: Is there an XSS risk here?

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
Binayak490-cyber: This loop queries every audited entity class in the DB - on a typical OpenMRS instance that can be 30+ queries per page load. As suggested and said by @wikumChamith, consider accepting a Set<Class<?>> parameter (the relevant field types from AuditServiceImpl) and only querying those specific classes instead of all of them. This would reduce queries from N down to only the few types actually referenced by the main entity.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
Binayak490-cyber: I want to follow up @druchniewicz on the isAssignableFrom() concern - fieldType.isAssignableFrom(actualType) only checks type compatibility, not actual data relationships. For example: if Obs has a Person field and a completely unrelated Person (different patient) is modified in the same revision during a bulk operation, it will appear in this Obs's Related Entities panel. Can we add a check that the related entity's actual ID matches a value referenced by the main entity, not just type compatibility?

omod/src/main/webapp/viewAudit.jsp
Binayak490-cyber: Yes, I think @wikumChamith, the <% %> scriptlet is still present. Even though it now correctly delegates to UtilClass.getEntityIdAsString(), the logic of resolving className, simpleName, and entityIdValue should be done inside ViewAuditController.java and passed as part of the model - similar to how relatedEntities is already handled. This keeps the JSP free of Java logic and makes it easier to test.

omod/src/main/webapp/viewAudit.jsp
Binayak490-cyber: Agreed. To add to the XSS concern - line 80 specifically uses <%= className %> directly inside the href query parameter without URL encoding. If the class name contains special characters or gets manipulated, this is an XSS risk. Please use <c:url> with proper parameter encoding or resolve and URL-encode the class name in ViewAuditController.java before passing it to the view.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
druchniewicz: this value is always set later so the initialization is redundant here

omod/src/main/webapp/viewAudit.jsp
druchniewicz: Corrected

omod/src/main/webapp/viewAudit.jsp
druchniewicz: Corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
druchniewicz: at this moment I don't see the case when it can occur because when we update something in the app it affects one patient and related entities, one encounter and related entities, one condition and related entities etc. But even if we can do somehow any bulk operation then I also think it is not a problem because we can see "what has been changed during this bulk operation/revision"

druchniewicz:
@wikumChamith I pushed new corrections and improvements according to your comments

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
Binayak490-cyber: Got it👍.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
Binayak490-cyber: Got it - understood that OpenMRS transactions are naturally scoped to one patient/encounter context. It would be helpful to add a short comment in the code explaining this assumption (e.g., // assumes revision entities are scoped to a single patient/encounter context) so future maintainers understand why type-based filtering is sufficient here. Otherwise it is okay.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
druchniewicz: corrected

druchniewicz:
@wikumChamith can you take a look at PR once again? Changes have been applied. Thanks in advance.

druchniewicz:
@wikumChamith this PR has been checked and adapted according to comments few times. Can we merge it and any possible future smaller fixes/improvements implement in new PRs?

dkayiwa:
@wikumChamith do you have any comments on the above?

wikumChamith:
> @wikumChamith this PR has been checked and adapted according to comments few times. Can we merge it and any possible future smaller fixes/improvements implement in new PRs?

We normally prefer to get PRs as close to perfect as possible before merging, because most of the time, follow-up PRs don't happen, and we're left with minor issues in the codebase. However, I think your PR is almost ready to be merged. I just have a few questions :)

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -321,6 +324,32 @@ public List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page
...
+
+        return allEntitiesInRevision.stream()
+                .filter(auditEntity -> {
+                    if (auditEntity.getEntity() == null) {
+                        return false;
+                    }
+
+                    Class<?> actualType = auditEntity.getEntity().getClass();
+                    for (Class<?> fieldType : fieldTypes.values()) {
+                        if (fieldType != null && fieldType.isAssignableFrom(actualType)) {
```
wikumChamith: Don't we already do the filter `isAssignableFrom(actualType)` in the DAO class?

```
        List<Class<?>> classesToQuery = getNonAbstractAuditedClasses().stream()
                .filter(c -> relevantClasses.stream().anyMatch(r -> r.isAssignableFrom(c)))
                .collect(Collectors.toList());
```

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
```diff
@@ -452,4 +453,51 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
...
+                        RevisionType revisionType = (RevisionType) array[2];
+
+                        if (entity != null) {
+                            Integer userId = revisionEntity != null ? revisionEntity.getChangedBy() : null;
+                            result.add(new AuditEntity<>(entity, revisionEntity, revisionType, userId));
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                log.warn("Could not find revision {} for entity {}: {}", revisionId, clazz.getSimpleName(), e.getMessage());
```
wikumChamith: It feels like catching all exceptions here could silently swallow real issues.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
druchniewicz: right indeed, I corrected it

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditDao.java
druchniewicz: I corrected it and I used existing approach using `isMissingAuditTableException(e)`

druchniewicz:
@wikumChamith I pushed commit with corrections to your comments from today

wikumChamith (approved):

PR #37, opened 2026-03-30
AUDIT-31: Fix null/empty/unreadable values in audit log field display

shubhangiisinghh:
## Problem
The audit log viewer was displaying raw "null" strings and blank values 
for changed fields. This happened because:
1. String.valueOf(null) produces the literal string "null" not a blank
2. OpenMRS domain object fields (Patient, User, Location etc.) triggered 
   Hibernate proxy loading which failed silently, showing nothing
3. If either old OR current value failed to read, BOTH were overwritten 
   with "Unable to read", losing valid data

- Replaced String.valueOf() with a null-safe toSafeString() helper
- Null field values now show [No value] instead of the literal string "null"
- OpenMRS domain objects (Patient, User, Location) now show Type(id=X)
  instead of triggering Hibernate proxy errors that caused blank values
- Old and current value failures are now handled independently so one
  broken field doesn't wipe out the other

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-31

- [x] Code style followed
- [x] Tests added (5 new unit tests in UtilClassUnitTest)
- [x] mvn clean package passed
- [x] All tests passed
- [x] Based on latest main branch

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
@@ -8,6 +8,22 @@
  */
 package org.openmrs.module.auditlogweb.api.utils;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneId;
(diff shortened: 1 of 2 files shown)


shubhangiisinghh:
@ManojLL @wikumChamith Would be much grateful if you have a look at this when you get a chance !!

PR #41, opened 2026-03-30
AUDIT-26: Exclude abstract classes from audited entity validation

shlok612:
### 🛠️ Description
This PR fixes AUDIT-26 by excluding abstract and base classes from the audited entity validation logic.

### ❌ Problem
Abstract classes like BaseOpenmrsData and BaseOpenmrsMetadata were being included in the entity selection dropdown. These classes are not audited, leading to NotAuditedException at runtime.

### ✅ Solution
Added a filter to exclude abstract classes using Modifier.isAbstract() during validation.

### 🔍 Changes
- Added filter to remove abstract classes from the entity list
- Prevents invalid entity selection and runtime errors

### 📌 Impact
- Fixes dropdown issue
- Prevents application crashes
- Ensures only valid audited entities are processed

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
@@ -32,11 +32,11 @@ import java.util.ArrayList;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-
 /**
  * Default implementation of the {@link AuditService} interface.
  *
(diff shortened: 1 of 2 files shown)


shubhangiisinghh:
Could we add a test case that confirms abstract classes are filtered out ?
There is no test added probably. A unit test that verifies abstract classes would make this PR complete I feel.

shlok612:
> Could we add a test case that confirms abstract classes are filtered out ? There is no test added probably. A unit test that verifies abstract classes would make this PR complete I feel.

Thanks for the suggestion! I’ll add a unit test to verify that abstract classes are excluded from entity validation.

PR #44, opened 2026-05-12
Audit-47 : Implement the Security Audit Logging

sudhanshu-raj:
Implemented the Security Audit Logging where we are capturing the events like  :  
```
    LOGIN_SUCCESS,
    LOGIN_FAILURE,
    ACCOUNT_LOCKED,
    LOGOUT,
    SESSION_TIMEOUT,
    PASSWORD_RESET_REQUEST_SUCCESS,
    PASSWORD_RESET_REQUEST_FAILURE,
    PASSWORD_RESET_SUCCESS,
    PASSWORD_RESET_FAILURE,
    PASSWORD_CHANGED_SUCCESS,
    PASSWORD_CHANGED_FAILURE,
    UNKNOWN;;
```
Capturing these above events using the AOP on the audit module like login success/failed & account locked security events and the password modify related events . HttpSessionListener for capturing session events. Storing the metadata using per request filter in a separate  thread local and session holder for storing and managing the in-application flow events. 

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-47

  No? -> execute above command


api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            auditService.logSecurityEvent(eventType, username, userId,ipAddress, userAgent, sessionId, details);
+
+            /* Checking whether this #changePassword call was triggered as part of a password reset
+              flow after successful secret answer verification. If so, then mark the reset request as completed and remove it.
+              This check is performed at the end because we first need to know whether
+              #changePassword belongs to a password reset or a normal password change,
+              so the correct event can be stored in the database.
+             */
```
sudhanshu-raj: Added raw comments for logic understanding , can remove it later.

ManojLL: move this `aop` stuff under api folder and and rename `advice` to `aop`

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
```diff
@@ -0,0 +1,134 @@
...
+    @Column(name = "user_agent", length = 500)
+    private String userAgent;
+
+    @Column(name = "session_id", length = 100)
+    private String sessionId;
+
+    @Column(name = "details", columnDefinition = "TEXT")
+    private String details;
+
+    public Long getId() {
```
wikumChamith: Can't we use class-level lombok `@Setter` and `@Getter`?

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditDetailController.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param model   model map for JSP
+     * @return ModelAndView pointing to viewSecurityAudit.jsp
+     */
+    @RequestMapping(method = RequestMethod.GET)
+    public ModelAndView showDetails(HttpServletRequest request, ModelMap model) {
+        try {
+
+            if(!Context.hasPrivilege(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)){
+                return new ModelAndView(ACCESS_DENIED_VIEW);
+            }
```
wikumChamith: Why are we doing a security check here? Don't they get handled at the service level?

sudhanshu-raj:
I removed the custom logout event publisher with with built in user session listener but for login attempt event there is extra event we are capturing which is "ACCOUNT_LOCKED" and that's why we can't use existing listener bcz it not capture this event.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
sudhanshu-raj: Done !

api/src/main/java/org/openmrs/module/auditlogweb/api/PasswordResetFlowContext.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+
+    private static final class PasswordResetFlowState {
+
+        private boolean resetCompleted;
+        private boolean isPasswordChangedBySystem;
+    }
+}
```
ManojLL:
```suggestion
}

```

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditSecurityEventType.java
```diff
@@ -0,0 +1,30 @@
...
+    PASSWORD_RESET_REQUEST,
+    PASSWORD_RESET;
+
+    public static AuditSecurityEventType fromName(String value) {
+        if (value == null || value.trim().isEmpty()) {
+            return null;
+        }
+        return AuditSecurityEventType.valueOf(value.trim().toUpperCase());
+    }
+}
```
ManojLL:
```suggestion
}

```

ManojLL (changes requested):

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
```diff
@@ -17,4 +17,5 @@ public final class AuditLogConstants {
 
     /* MODULE PRIVILEGES */
     public static final String VIEW_AUDIT_LOGS = "View Audit Logs";
+    public static final String VIEW_SECURITY_AUDIT_LOGS = "View Security Audit Logs";
```
wikumChamith: We need to register these privileges in  config.xml

omod/src/main/webapp/securityauditlogs.jsp
```diff
@@ -0,0 +1,159 @@
...
+--%>
+<%@ include file="/WEB-INF/template/include.jsp"%>
+<%@ include file="/WEB-INF/template/header.jsp"%>
+<%@ include file="localHeader.jsp"%>
+
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@300;400;600&display=swap">
+<link rel="stylesheet" type="text/css" href="<openmrs:contextPath/>/moduleResources/auditlogweb/css/securityauditlogs.css" />
+
+
+<div class="security-page">
```
wikumChamith: Some fields here are rendered with raw `${...}` instead of `<c:out>`, for example `event.username`, `event.eventType`, and the echoed filter values. Wrap these in `<c:out>` (or `fn:escapeXml` for attribute values) like you already did for `sessionId` and `details`.

omod/src/main/webapp/viewSecurityAudit.jsp
```diff
@@ -0,0 +1,188 @@
...
+--%>
+<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>
+<%@ include file="/WEB-INF/template/include.jsp"%>
+<%@ include file="/WEB-INF/template/header.jsp"%>
+<%@ include file="localHeader.jsp"%>
+
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@300;400;600&display=swap">
+<link rel="stylesheet" type="text/css" href="<openmrs:contextPath/>/moduleResources/auditlogweb/css/viewSecurityAudit.css" />
+
+<div class="security-audit-detail-container">
```
wikumChamith: Same XSS issus in here

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,154 @@
...
+@RequiredArgsConstructor
+public class AuthenticationAdvice {
+
+    private final static Logger log = LoggerFactory.getLogger(AuthenticationAdvice.class);
+    private final AuditService auditService;
+    private final SessionFactory sessionFactory;
+
+
+    @Around("execution(* org.openmrs.api.db.hibernate.HibernateContextDAO.authenticate(..))")
+    public Object authenticate(ProceedingJoinPoint joinPoint) throws Throwable {
```
wikumChamith: The three `logSecurityEvent()` calls here aren't wrapped in try/catch like they are in `PasswordAuditAdvice` and the listeners. If the insert fails, the success path breaks login for everyone, and in the failure paths the new exception replaces the original `ContextAuthenticationException`. Please wrap these calls so an audit failure is logged at error level but never propagates.

wikumChamith:
@sudhanshu-raj I noticed that some of the files are missing our license header, please add them.

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
sudhanshu-raj: Done !

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditDetailController.java
sudhanshu-raj: yeah, we already checking it . Removed it now.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
sudhanshu-raj: Done !

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
sudhanshu-raj: This is great catch, now wrapped the `logSecurityEvent()` inside try/catch and added test cases to prove this.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditDetailController.java
```diff
@@ -0,0 +1,98 @@
...
+                relatedEvents = auditService.getRelatedSecurityEvents(event.getSessionId(), RELATED_EVENTS_LIMIT);
+            }
+
+            model.addAttribute("event", event);
+            model.addAttribute("relatedEvents", relatedEvents != null ? relatedEvents : new java.util.ArrayList<>());
+            model.addAttribute("page", "securityauditlogs");
+
+            return new ModelAndView(VIEW, model);
+
+        } catch (Exception e) {
```
wikumChamith: Won't this also catch the `APIAuthenticationException` thrown by `@Authorized` and leak the privilege name instead of showing the accessDenied.jsp?

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
```diff
@@ -0,0 +1,67 @@
+    /**
```
wikumChamith: Do we need this leading whitespace?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,171 @@
...
+            throw ex;
+        }
+    }
+
+    private void safelyLogSecurityEvent(AuditSecurityEventType eventType, String username, Integer userId,
+                                        String ipAddress, String userAgent, String sessionId, String detailsJson) {
+        try {
+            auditService.logSecurityEvent(eventType, username, userId, ipAddress, userAgent, sessionId, detailsJson);
+        } catch (Exception e) {
+            log.error("Failed to log authentication security event", e);
```
wikumChamith:
```suggestion
            log.error("Failed to log authentication security event [{}] for user [{}]", eventType, username, e)
```

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
```diff
@@ -269,4 +270,64 @@ List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page, int s
...
+         *
+         * @param eventType optional event type filter
+         * @param username optional username filter (partial, case-insensitive)
+         * @param startDate optional inclusive start time filter
+         * @param endDate optional inclusive end time filter
+         * @param page zero-based page index
+         * @param size page size
+         * @return paginated list of matching security events
+         */
+        List<AuditSecurityEvent> getSecurityEvents(String eventType, String username,
```
wikumChamith: Don't we need `@Authorized` for this?

omod/src/main/webapp/securityauditlogs.jsp
sudhanshu-raj: fixed

omod/src/main/webapp/viewSecurityAudit.jsp
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
sudhanshu-raj: Done !

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
sudhanshu-raj: Removed !

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditDetailController.java
sudhanshu-raj: Fixed it.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * It listens for activity or specifically the password related functions like #isSecretAnswer & #changePassword from the OpenMRS core,
+ * and stamps the event like : PASSWORD_CHANGED, PASSWORD_RESET_REQUEST, PASSWORD_RESET
+ *
+ */
+@Aspect
+@Component
+@RequiredArgsConstructor
+
```
wikumChamith:
```suggestion
```

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
```diff
@@ -269,4 +270,66 @@ List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page, int s
...
+     *                    SESSION_TIMEOUT, PASSWORD_CHANGED, PASSWORD_RESET
+     * @param username    the username involved in the event (may be null for anonymous sessions)
+     * @param userId      the OpenMRS user_id (nullable when the user doesn't exist in the DB)
+     * @param ipAddress   the client IP address extracted from the request (nullable)
+     * @param userAgent   the HTTP User-Agent header value (nullable)
+     * @param sessionId   the HTTP session ID (nullable)
+     * @param detailsJson optional JSON string with additional context (nullable)
+     */
+    @Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
+    void logSecurityEvent(AuditSecurityEventType eventType, String username, Integer userId,
```
wikumChamith: @sudhanshu-raj have you tested this PR? I'm seeing a privilege error when trying to record these audits.

`org.openmrs.api.APIAuthenticationException: Privileges required: View Security Audit Logs`

This is because you've annotated the `logSecurityEvent` method with `@Authorized`. We need to either remove the `@Authorized` from `logSecurityEvent` since it's an internal write path, or wrap the call with a proxy privilege.

sudhanshu-raj: Correct, it should not here, i accidentally added on last commit.

wikumChamith: It's really important to manually test your changes. Please make sure you always do that.

sudhanshu-raj: My bad,  Sure I will check these things before .

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,171 @@
...
+                        sessionId,
+                        buildLoginDetails("INVALID_USERNAME",false));
+                throw ex;
+            }
+
+            String message = ex.getMessage();
+            AuditSecurityEventType eventType = null;
+            String reason = null;
+            boolean isAccountLocked = false;
+            if("Invalid number of connection attempts. Please try again later.".equals(message)){
```
wikumChamith: What happens if this message gets changed in the core?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
```diff
@@ -0,0 +1,188 @@
...
+              flow after successful secret answer verification. If so, then mark the reset request as completed and remove it.
+              This check is performed at the end because we first need to know whether
+              #changePassword belongs to a password reset or a normal password change,
+              so the correct event can be stored in the database.
+             */
+            if ("changePassword".equals(methodName) && PasswordResetFlowContext.hasPendingResetRequest(sessionId)
+                    && PasswordResetFlowContext.isPasswordChangedBySystem(sessionId)) {
+                PasswordResetFlowContext.markResetCompleted(sessionId);
+            }
+
```
wikumChamith: If `logSecurityEvent` throws here, `markResetCompleted` never runs, and the session state is stranded as pending.  The next normal password change on this session then gets logged as `PASSWORD_RESET`. Can we decide the event type before logging and move the cleanup into a finally?

api/src/main/java/org/openmrs/module/auditlogweb/api/PasswordResetFlowContext.java
```diff
@@ -0,0 +1,129 @@
...
+            state.isPasswordChangedBySystem = false;
+        }
+    }
+
+    /**
+     * Marks the password reset flow for the given session as completed and then removes it.
+     *
+     * @param sessionId the session identifier for the password reset flow
+     */
+    public static void markResetCompleted(String sessionId) {
```
wikumChamith: This sets `resetCompleted = true` and then immediately removes the entry, so the flag is never actually read by anyone. Also, `computeIfAbsent` can recreate a just-removed entry, racing a concurrent `markResetRequest`. Can we just do `STATES.remove(sessionId)` here? The `resetCompleted` field then becomes dead and can be dropped from `PasswordResetFlowState`.

sudhanshu-raj: yeah correct approach,  removing it directly now.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
sudhanshu-raj: Good edge case caught,  now i made sure,  reset flow is updated in any case.

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
sudhanshu-raj, line 1: I have updated this entity, please drop this existing table and liquibase changelog before testing again.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
sudhanshu-raj: I though it would work but not seems to standard way, so I now checking user `USER_PROPERTY_LOCKOUT_TIMESTAMP` property then acts and log the account lockout and login failure events. Also added new test cases around it.

api/src/main/java/org/openmrs/module/auditlogweb/api/PasswordResetFlowContext.java
```diff
@@ -0,0 +1,123 @@
...
+            state.isPasswordChangedBySystem = false;
+        }
+    }
+
+    /**
+     * Marks the password reset flow for the given session as completed by removing its state.
+     *
+     * @param sessionId the session identifier for the password reset flow
+     */
+    public static void markResetCompleted(String sessionId) {
```
wikumChamith: `markResetCompleted` is the only thing that removes entries, so an abandoned reset (user verifies the secret answer but never finishes) leaks its entry forever. Since `isSecretAnswer` is pre-auth, this can grow unbounded. I think we should add a `clear(sessionId)` method and call it from `SessionTimeoutListener.sessionDestroyed` (before the early returns), so every destroyed session reclaims its state.

wikumChamith (commented):
Can we get tests for the classes in the web layer?

api/src/main/java/org/openmrs/module/auditlogweb/api/PasswordResetFlowContext.java
sudhanshu-raj: Great finding, so yeah it can be lead to memory leak . Fixed now !

sudhanshu-raj:
> Can we get tests for the classes in the web layer?

Added the test cases for security web controller and context filter classes.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditSecurityEventType.java
```diff
@@ -0,0 +1,30 @@
...
+    SESSION_TIMEOUT,
+    PASSWORD_CHANGED,
+    PASSWORD_RESET_REQUEST,
+    PASSWORD_RESET;
+
+    public static AuditSecurityEventType fromName(String value) {
+        if (value == null || value.trim().isEmpty()) {
+            return null;
+        }
+        return AuditSecurityEventType.valueOf(value.trim().toUpperCase());
```
wikumChamith: What will happen if someone passes a value that isn't in the enum?

wikumChamith:
Can we add a test to the [[ServiceAuthorizationTest](https://github.com/openmrs/openmrs-module-auditlogweb/blob/86389b46acacbfd542f85b1418ea22c7d4422859/api/src/test/java/org/openmrs/module/auditlogweb/api/ServiceAuthorizationTest.java#L22)](https://github.com/openmrs/openmrs-module-auditlogweb/blob/86389b46acacbfd542f85b1418ea22c7d4422859/api/src/test/java/org/openmrs/module/auditlogweb/api/ServiceAuthorizationTest.java#L22) for `VIEW_SECURITY_AUDIT_LOGS`?

omod/src/test/java/org/openmrs/module/auditlogweb/web/AuditContextFilterTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class AuditContextFilterTest {
```
wikumChamith, line 36: Can we add a test that makes buildContext throw and asserts doFilter still ran and SecurityAuditContext.get() is null?  Letting that exception propagate would break every request silently.

sudhanshu-raj: Added the test case for that it should not effect the rest application if we got anything wrong during build context.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditSecurityEventType.java
sudhanshu-raj: yeah this can hurt the flow, fixed it now if someone pass invalid event type then will return empty list

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
```diff
@@ -0,0 +1,191 @@
...
+               confirms that user has requested for password reset request and just few milliseconds later system has reset with temporary pass.
+             */
+            if("changePassword".equals(methodName) && PasswordResetFlowContext.hasPendingResetRequest(sessionId)
+            && !PasswordResetFlowContext.isPasswordChangedBySystem(sessionId)) {
+                PasswordResetFlowContext.setPasswordChangedBySystem(sessionId,true);
+                return;
+            }
+
+            // Reached here means the request is only for verifying the secret answer before allowing the password reset.
+            if ("isSecretAnswer".equals(methodName)) {
```
wikumChamith: `markResetRequest(sessionId)` runs for every `isSecretAnswer` call, even when the answer is wrong. That leaves a stale "pending reset" flag on the session, and `AuthenticationAdvice.java:58` then skips the `LOGIN_SUCCESS` audit when the user logs in normally in that same session. So a wrong security answer can cause us to lose a real login event.

I think we should only call `markResetRequest` when `Boolean.TRUE.equals(returnValue)`.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
```diff
@@ -0,0 +1,191 @@
...
+ */
+@Aspect
+@Component
+@RequiredArgsConstructor
+public class PasswordAuditAdvice  {
+
+    private static final Logger log = LoggerFactory.getLogger(PasswordAuditAdvice.class);
+    private final AuditService auditService;
+
+    @AfterReturning(
```
wikumChamith: The pointcut watches `UserService.changePassword(..)`, but the secret answer reset flow in OpenMRS 2.7.0 actually finishes through `changePasswordUsingSecretAnswer(..)` and `changePasswordUsingActivationKey(..)`. So the `PASSWORD_RESET` branch and the temp password skip never trigger in real usage. The tests pass only because they mock a `changePassword` method that the real flow doesn't call.

I think we need to add the real reset methods to the pointcut, and add an integration test against the actual flow.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -318,6 +322,28 @@ public List<AuditEntity<?>> getAllRevisionsAcrossEntitiesWithEntityType(int page
...
      * Counts audit logs across entities with filtering.
      */
+    /**
+     * Builds an {@link AuditSecurityEvent}, stamps the current timestamp, and delegates
+     * persistence to the DAO. Null-safe: all optional parameters are accepted without
+     * any non-null constraint here.
+     */
+    @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void logSecurityEvent(AuditSecurityEventType eventType, String username, Integer userId,
```
wikumChamith: save() only queues the insert. The actual write happens at commit, which is after logSecurityEvent returns, so it's outside the try/catch in safelyLogSecurityEvent. If a constraint or length limit is hit (null eventType, oversized username/user_agent), the event is lost and nothing logs it. For an audit feature, losing events quietly is the worst case. What if we call `session.flush()` inside logSecurityEvent so errors surface where we can catch and log them, and trim/validate fields to their column limits first.

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Date;
+
+/**
+ * Hibernate entity representing a single security-related audit event. 
+ */
+@Entity
+@Table(name = "audit_security_event")
+@Getter
+@Setter
+public class AuditSecurityEvent extends BaseOpenmrsObject {
```
wikumChamith, line 40: `@Getter @Setter` makes every field mutable, including the id. An audit row should be write-once, but right now anyone can change `eventType`, `eventTime`, or `userId` after it's created. The not-null rules only live in the DB, so a missing value blows up as a Hibernate error at flush instead of failing early.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
sudhanshu-raj: Or we can take one extra parameter on `PasswordResetFlowContext` name `isSecretAnswerVerified ` to check if this value is true then only ignore the login event bcz this can only happen after success secret answer verified which results in login success so it ignores in this case and rest will be as it is. Bcz if bad actors continuously try to do password reset request  and if we `markResetRequest `  only on `Boolean.TRUE.equals(returnValue)` then we will miss those failed event(password_reset_request) logs, that's why need extra param.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
sudhanshu-raj: I only took the flow for legacy-ui module([here](https://github.com/openmrs/openmrs-module-legacyui/blob/7a8b83603aefa83633209eeff70f72413a63fb9e/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java#L169))  in this case, instead should also take REST API  cases which can directly call those methods you mentioned too so yeah we need to add these methods  `changePasswordUsingSecretAnswer(..)` `changePasswordUsingActivationKey(..)`  under the pointcut list and here we can directly log the `PASSWORD_RESET` event instead of `PASSWORD_RESET_REQUEST` too.

sudhanshu-raj: Fixed it

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PasswordAuditAdvice.java
sudhanshu-raj: Done

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
sudhanshu-raj: I though of this will happen in one go and synchronously and under the safelyLogSecurityEvent stack but there can be cases where it can bypass transaction proxy an will commit later so yeah this is good edge case. I extend the columns limit to standard industry size and trim if they exceeds limit and most important flushing the action immediate . Please  delete the existing liquibase changlelog for AuditSecurityEvent.

api/src/main/java/org/openmrs/module/auditlogweb/AuditSecurityEvent.java
sudhanshu-raj: Made the fields immutable and pre null checks/

wikumChamith:
image

Could you tell me why the username isn't being displayed? Also, I think it would be better if we could replace the user ID with the UUID

sudhanshu-raj:
Username is not displaying here probably because user has not set the username , but there is bug on explorer page there it shows the username as system for same audit log ,  i will fix this . And yeah we can use user UUID instead of user ID

sudhanshu-raj:
@wikumChamith  , can we put the system id if username not there  in username column?

wikumChamith:
> @wikumChamith , can we put the system id if username not there in username column?

Yeah

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditlogwebController.java
```diff
@@ -0,0 +1,104 @@
...
+            @RequestParam(value = "eventType", required = false) String eventType,
+            @RequestParam(value = "username", required = false) String username,
+            @RequestParam(value = "startDate", required = false) String startDate,
+            @RequestParam(value = "endDate", required = false) String endDate,
+            @RequestParam(value = "page", defaultValue = "0") int page,
+            @RequestParam(value = "size", defaultValue = "15") int size,
+            Model model) {
+
+        Date start = UtilClass.parseDate(startDate, false);
+        Date end = UtilClass.parseDate(endDate, true);
```
wikumChamith:
```suggestion
        Date start = UtilClass.toStartDate(UtilClass.parse(startDate));
        Date end = UtilClass.toEndDate(UtilClass.parse(endDate));
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+    private void safelyLogSecurityEvent(AuditSecurityEventType eventType, String username, String userUuid,
+                                        String ipAddress, String userAgent, String sessionId, String detailsJson) {
+        try {
+            auditService.logSecurityEvent(eventType, username, userUuid, ipAddress, userAgent, sessionId, detailsJson);
+        } catch (Exception e) {
+            log.error("Failed to log authentication security event [{}] for user [{}]", eventType, username, e);
+        }
+    }
+
+    private User resolveUser(String login){
```
wikumChamith:
```suggestion
    private User resolveUser(String login) {
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+                            User.class)
+                    .setParameter(1, login).setParameter(2, login).setParameter(3, loginWithDash).uniqueResult();
+        }
+        catch (Exception e) {
+            log.error("Error while resolving user for audit");
+        }
+        return loginUser;
+    }
+
+    private String buildLoginDetails(String reason,boolean isAccountLocked) {
```
wikumChamith:
```suggestion
    private String buildLoginDetails(String reason, boolean isAccountLocked) {
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+
+            Object[] args = joinPoint.getArgs();
+            String login = null;
+
+            if (args != null && args.length > 0) {
+                login = (String) args[0];
+            }
+
+            User user = resolveUser(login);
+            if(user == null){
```
wikumChamith: When the user-lookup query throws here (a DB/session error or something), `resolveUser` returns null, and the caller then records the event as `INVALID_USERNAME`. So a real account whose lookup just failed gets logged as "no such user."

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+// It logs the event related to authentication like login success/failed and account locked.
+@Aspect
+@Component
+@RequiredArgsConstructor
+public class AuthenticationAdvice {
+
+    private final static Logger log = LoggerFactory.getLogger(AuthenticationAdvice.class);
```
wikumChamith:
```suggestion
    private static final Logger log = LoggerFactory.getLogger(AuthenticationAdvice.class);
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+                    userAgent,
+                    sessionId,
+                    "");
+
+            // Marks current pre-fixation session id so SessionTimeoutListener ignores it.
+            // The login flow invalidates that session later during fixation protection.
+            markSessionAsLoginFixation();
+
+            return result;
+        } catch(ContextAuthenticationException ex){
```
wikumChamith:
```suggestion
        } catch (ContextAuthenticationException ex){
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+
+            AuditSecurityEventType eventType = null;
+            String reason = null;
+            boolean isAccountLocked = false;
+            if (isAccountLocked(user)) {
+                log.debug("Authentication event : ACCOUNT_LOCKED");
+                eventType = AuditSecurityEventType.ACCOUNT_LOCKED;
+                isAccountLocked = true;
+                reason = "Too many attempts";
+            }else{
```
wikumChamith:
```suggestion
            }
            else {
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -0,0 +1,210 @@
...
+    public Object authenticate(ProceedingJoinPoint joinPoint) throws Throwable {
+
+        AuditLogContext ctx = AuditLogContext.get();
+        String sessionId = ctx != null ? ctx.getSessionId() : null;
+        String ipAddress = ctx != null ? ctx.getIpAddress() : null;
+        String userAgent = ctx != null ? ctx.getUserAgent() : null;
+
+        try{
+            Object result = joinPoint.proceed();
+            if(!(result instanceof User)) return result;
```
wikumChamith: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Code-Style

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
sudhanshu-raj: That's good case, here we can like check if the fetched username is null from the DB that means user not found and so the `INVALID_USERNAME` and if there is any DB errors or anything comes then will abort and re-throw the exception

wikumChamith (approved):

ManojLL:
@sudhanshu-raj could you please resolve the merge conflicts?

sudhanshu-raj:
> @sudhanshu-raj could you please resolve the merge conflicts?

done!

ManojLL (approved):
LGTM

PR #52, opened 2026-06-26
Audit 53 : Implement Read Audit Logging

sudhanshu-raj:
We are implementing the Read Audit Logging, **about this feature** : Whenever any user view any record on OpenMRS like getting records for patients, obs, encounters, addresses etc. then this feature will log those activity in our audit tables so that we will know who viewed what and when and this is also important as compliance perspective in health care domain like HIPAA and others.

For this we  gonna use AOP and add pointcuts around the most relevant entities get method so that we can intercept those read calls from our audit module. We gonna store the entity UUID as a list with it's entity type  as a entity metadata to know who viewed what.  And we also storing the user device metadata like IP Address, User Agent, session id along with the user's username and their UUID. 

To save the DB explosion from same duplicate  records under a time frame of like 2 or 5 min, I'm using the caching to know whether the same record we stored few minutes earlier or not and keeping the TTL for 2 min for now. 

And to not put the load on original fetch request during our auditing, I'm using `BlockingQueue` to store the requests and then consuming it on background thread  and also draining it like suppose there is 50 requests on queue then save those all at one in a batch instead of one by one thus it will save DB round up time and pool. 

I have talked more about this on my GSoC weekly blog  [here](https://medium.com/@rajsudhanshu/gsoc-week-5-developing-the-read-audit-logging-feature-5a721e7ae9da) .

## Issue I worked on
see https://openmrs.atlassian.net/browse/AUDIT-53

  No? -> execute above command


api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PatientDataReadAOP.java
```diff
@@ -0,0 +1,178 @@
...
+			returnDataType = getMethodReturnDataType(method);
+			
+			AuditLogContext ctx = AuditLogContext.get();
+			if (ctx != null) {
+				username = ctx.getLoggedInUsername() != null ? ctx.getLoggedInUsername() : null;
+				userUUID = ctx.getLoggedInUserUUID() != null ? ctx.getLoggedInUserUUID() : null;
+				ipAddress = ctx.getIpAddress() != null ? ctx.getIpAddress() : null;
+				userAgent = ctx.getUserAgent() != null ? ctx.getUserAgent() : null;
+				sessionId = ctx.getSessionId() != null ? ctx.getSessionId() : null;
+			}
```
wikumChamith: `AuditLogContext` is only populated on web requests, so it's null for daemon reads and non-web human access. Should we fall back to `Context.getAuthenticatedUser()` here, attributing real human reads and skipping daemon/unauthenticated ones, instead of building a log with a null user? This is also what causes the `NOT NULL user_uuid` insert to fail and get dropped.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,111 @@
...
+		log.info("Stopping ReadAuditWorker background thread...");
+		running = false;
+		if (workerThread != null) {
+			workerThread.interrupt();
+		}
+	}
+	
+	public void submitTask(ReadAuditLog readAuditLog) {
+		boolean isAdded = queue.offer(readAuditLog);
+		if (!isAdded) {
```
wikumChamith, line 86: When the queue is full, `offer()` returns false, and we only log it. Isn't this a silent loss of audit records exactly under load? For a compliance trail, should a dropped read really just be a log line?

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,111 @@
...
+			log.error("Failed to save read audit logs in batch, falling back to one-by-one save", e);
+			for (ReadAuditLog logEntry : batch) {
+				try {
+					readAuditService.logReadAudit(logEntry);
+				}
+				catch (Exception ex) {
+					log.error("Failed to save individual read audit log in fallback", ex);
+				}
+			}
+		}
```
wikumChamith: After the batch save fails, the fallback reuses the same (likely broken) Hibernate session, so won't every one-by-one retry also fail? Should each fallback record get a fresh session so one bad row doesn't sink the other 49?

sudhanshu-raj: yeah, true it can be effect other save also that's why now opening the new session after each fail

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: This is one of good case, yeah we have to bring some standard design solution for this, like 
1. We can stop the caller AOP thread from submitting the task into queue instead will excecute the save audit log work on their thread but it can make those requests slightly slow as we already will have probably lot of database load running .  But this will be temporary and as we have sudden spike may be we can trade off this with slight slow request.
2. Another is we can increase the backgrounds threads make that request save and drain the queue fast.

Else what's your though on this ?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PatientDataReadAOP.java
sudhanshu-raj: true, we should also look for auditing the non-web human requests, fixed now

sudhanshu-raj:
Hi @ManojLL @wikumChamith , this got ready, now the review time  starts 😅

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
```diff
@@ -0,0 +1,93 @@
...
+			log.error("Failed to load security audit logs", e);
+			model.addAttribute("errorMessage", "An error occurred while loading security audit logs.");
+			model.addAttribute("readAuditLogs", Collections.emptyList());
+			
+		}
+		return new ModelAndView(VIEW);
+	}
+	
+	private List<String> getEntityTypes() {
+		return Arrays.asList("PATIENT");
```
dkayiwa: This needs fixing before merge: the dropdown only ever offers PATIENT, while the aspect audits 16 services, so none of the other logged types (Encounter, Obs, Concept, User, ...) can be filtered from the UI. It also doesn't match what we actually store. entityName holds Java simple names from getMethodReturnDataType(), like "Patient", so the uppercase "PATIENT" only matches thanks to MySQL's case-insensitive collation and returns zero rows on a case-sensitive one. I'd derive the options from the data (select distinct entityName, cheap and always accurate) instead of hand-maintaining a list here.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditAOP.java
```diff
@@ -0,0 +1,189 @@
...
+		return readAuditService.auditReadRequest(joinPoint);
+	}
+	
+	@Around("execution(* org.openmrs.api.OrderService.getOrder*(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.getDiscontinuationOrder(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.getRevisionOrder(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.getAllOrders*(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.getActiveOrders(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.getCareSetting*(..)) || "
+	        + "execution(* org.openmrs.api.OrderService.discontinueOrder(..)) || "
```
dkayiwa: Is discontinueOrder here on purpose? It creates the discontinuation order, so it's a write, not a read. If it stays, every discontinuation shows up in the read trail as a successful Order read (on top of the envers write audit), which will mislead anyone reconstructing who viewed what. I'd drop it from the pointcut:
```suggestion
```

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditAOP.java
```diff
@@ -0,0 +1,189 @@
...
+	
+	@Around("execution(* org.openmrs.api.CohortService.getCohort*(..)) || "
+	        + "execution(* org.openmrs.api.CohortService.getAllCohorts(..))")
+	public Object auditCohortDataRead(ProceedingJoinPoint joinPoint) throws Throwable {
+		return readAuditService.auditReadRequest(joinPoint);
+	}
+	
+	@Around("(execution(* org.openmrs.api.ConceptService.getConcept*(..)) || "
+	        + "execution(* org.openmrs.api.ConceptService.getDrug*(..)) || "
+	        + "execution(* org.openmrs.api.ConceptService.getAllConcept*(..)) || "
```
dkayiwa: A scope question on auditing ConceptService and the other pure-metadata services (Form, Location, Provider, User): concept reads are the hottest path in the system (every form render and dictionary page fans out into them) and they aren't patient data. With the 2 minute cache TTL, each distinct concept per user and IP becomes a fresh log row plus metadata row every couple of minutes, and a single getAllConcepts() on a CIEL-sized dictionary writes tens of thousands of read_audit_entity_metadata rows in one call while blowing past the 10k cache cap, so the next call re-logs most of them. Was auditing terminology reads part of the AUDIT-53 scope, or could this start with patient-linked entities only (Patient, Encounter, Obs, Order, Visit, Condition, Diagnosis, Cohort, MedicationDispense), with metadata services opt-in later? Capping targets per event would also bound the worst case. My vote is patient-linked only for the first cut.

api/src/main/resources/liquibase.xml
```diff
@@ -45,4 +45,69 @@
         </createTable>
     </changeSet>
 
+    <changeSet id="auditlogweb-20260622-create-read-audit-log-table" author="sudhanshu_raj">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="read_audit_log"/>
+            </not>
+        </preConditions>
+        <createTable tableName="read_audit_log">
```
dkayiwa, line 54: Optional, but cheapest to do while the table is new: read_audit_log grows at per-read rate, the explorer sorts every page by event_time, and getRelatedReadLogs filters on session_id, both unindexed, so those queries become full table scans once there's real volume. Worth adding createIndex changesets for event_time and session_id (and possibly entity_name) while this PR is still open.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
```diff
@@ -0,0 +1,93 @@
...
+			model.addAttribute("endDate", endDate);
+			model.addAttribute("page", "readauditlogs");
+			model.addAttribute("entityTypes", getEntityTypes());
+		}
+		catch (APIAuthenticationException e) {
+			return new ModelAndView(ACCESS_DENIED_VIEW, model);
+		}
+		catch (Exception e) {
+			log.error("Failed to load security audit logs", e);
+			model.addAttribute("errorMessage", "An error occurred while loading security audit logs.");
```
dkayiwa: Take or leave: a few strings on the read pages still say "security", copied from the security screens. The user-visible error here and the log line above both say "security audit logs", the empty-table row in readAuditLogs.jsp says "No security audit logs found for the given criteria.", and ReadAuditDetailWebController logs "Error loading security audit detail". Confusing for whoever reads the page or greps the logs.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PatientDataReadAOP.java
dkayiwa: The fallback covers the ctx == null case, but I think a web-request variant of the same gap survives. AuditContextFilter sets a context for every request and captures the user only if authentication already happened when the filter ran. For stateless API clients (webservices.rest Basic auth authenticates inside its own module filter, and module filter order follows module start order) the audit filter can run first, so the context reaches saveReadAuditRequest() non-null with a null user UUID and the read is skipped. Net effect: REST reads with per-request auth are never audited, and since it depends on module start order it can work on one install and silently gap on another. Because the advice runs at service-call time, after authentication has completed, extending the fallback to also apply when ctx.getLoggedInUserUUID() is null (keeping the IP, user agent and session from ctx) should close it. Worth a unit test covering a context set with no user.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
dkayiwa: One more wrinkle for whichever direction you pick here: the dedup cache is marked before the log is durably enqueued (ReadAuditServiceImpl sets the key, then offer() may drop the task), so under exactly the overload this thread is about, a dropped read is also suppressed for the next 2 minutes, since repeats hit the cache and never re-attempt. Setting the cache keys only after a successful offer() would keep a full queue from also poisoning the dedup window.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/PatientDataReadAOP.java
sudhanshu-raj: yeah true, I was using not soo best way to handle this, since saveReadAuditRequest() gonna execute after the authentication only so we can check if userUUID is null then probably the audit context filter has run before the webservice  filter chain so now I am checking the userUUID again before saving the audit and as this time user must be authenticated we can got the user details and hence can saved it successfully. Added new test cases to cover this scenario too.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: yeah, adding the new key into the cache only after the task got added to queue to prevent silent ignore of audits if there is overload, btw fixed it now

api/src/main/resources/liquibase.xml
sudhanshu-raj: yeah, we can do this now

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
sudhanshu-raj: Ahh, I missed this,  my bad.  New approach seems better way, I have also added indexing on entity_name and by default if there is empty list then it will return Patient to the dropdown .

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditAOP.java
sudhanshu-raj: yeah it should not be here, I'd accidentally added it.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
sudhanshu-raj: Ohh, I should be careful while pasting this. Fixed !

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditAOP.java
sudhanshu-raj: yeah, I guess we can lean more towards the patient-linked entities.

++ @wikumChamith @ManojLL

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: > This is one of good case, yeah we have to bring some standard design solution for this, like
> 
> 1. We can stop the caller AOP thread from submitting the task into queue instead will excecute the save audit log work on their thread but it can make those requests slightly slow as we already will have probably lot of database load running .  But this will be temporary and as we have sudden spike may be we can trade off this with slight slow request.
> 2. Another is we can increase the backgrounds threads make that request save and drain the queue fast.
> 
> Else what's your though on this ?

I implemented the first approach here

api/src/main/resources/liquibase.xml
```diff
@@ -45,4 +45,103 @@
...
+    <changeSet id="auditlogweb-20260704-create-read-audit-entity-name-idx" author="sudhanshu_raj">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="read_audit_log_entity_name_idx"/>
+            </not>
+        </preConditions>
+        <createIndex tableName="read_audit_log" indexName="read_read_audit_log_entity_name_idx">
+            <column name="entity_name"/>
+        </createIndex>
+    </changeSet>
```
wikumChamith, line 146: The precondition checks for `read_audit_log_entity_name_idx`, but `createIndex` creates `read_read_audit_log_entity_name_idx`, so the index name has a typo, and the "already exists" guard never matches what it creates.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/ReadAuditServiceImpl.java
```diff
@@ -0,0 +1,267 @@
...
+						}
+						userUUID = user.getUuid();
+					}
+				}
+			}
+			
+			if (userUUID == null) {
+				return;
+			}
+
```
wikumChamith: This early-returns with no log when the user can't be resolved, and it sits before the `!isReadSuccess` submit, so a failed/unauthenticated read attempt is silently dropped.

api/src/main/java/org/openmrs/module/auditlogweb/api/ReadAuditService.java
```diff
@@ -0,0 +1,46 @@
...
+	
+	@Authorized(AuditLogConstants.VIEW_READ_AUDIT_LOGS)
+	List<ReadAuditLog> getRelatedReadLogs(String sessionId, int limit);
+	
+	@Authorized(AuditLogConstants.VIEW_READ_AUDIT_LOGS)
+	List<String> getEntityTypes();
+	
+	Object auditReadRequest(ProceedingJoinPoint joinPoint) throws Throwable;
+	
+	void saveReadAuditRequest(String entityName, boolean isReadSuccess, Object result);
```
wikumChamith: Any caller can call `Context.getService(ReadAuditService.class).saveReadAuditRequest(...)` and forge audit entries. The read methods are all `@Authorized`. Should these even be on the public interface, or should they move behind a non-interface component that the aspect delegates to?

api/src/main/resources/liquibase.xml
sudhanshu-raj: fixed!

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/ReadAuditServiceImpl.java
sudhanshu-raj: yeah, right it can be case where un-auth user can hit get/fetch methods, i have marked them as anonymous user

api/src/main/java/org/openmrs/module/auditlogweb/api/ReadAuditService.java
sudhanshu-raj: True, these method need to protected so that not get called by the elsewhere, I moved those helper method to it's own component class and now it can get call by AOP package classes only.

Here as this got hit, it also point towards the other methods like `logReadAudits` and `logReadAudits`  which is also has same condition, I guess we will also secure this.

dkayiwa: The `logReadAudit`/`logReadAudits` pair here is still reachable without authorization. `ReadAuditServiceImpl` implements `ReadAuditWriteService`, and the registered `auditlogweb.ReadAuditService` proxy sets no `proxyInterfaces`, so `TransactionProxyFactoryBean` builds it over every interface the target implements (spring-aop's `AbstractSingletonProxyFactoryBean` falls back to `getAllInterfacesForClass`; I checked the 5.3 source). So `((ReadAuditWriteService) Context.getService(ReadAuditService.class)).logReadAudit(forged)` reaches the persist path with no `@Authorized` and no VIEW privilege. If it merges like this, any code holding a Context reference can inject or forge read-audit rows, and for a compliance trail that tamper-resistance is the one property that has to hold.

`@Authorized` alone won't close it, since the background worker calls these with no authenticated user. I'd pull the write impl into its own `@Component` that's kept out of the `moduleService` list (and off the registered proxy), so it stays callable only by the aspect and the worker, with a test asserting an unprivileged `Context.getService(...)` cast can't persist a row.

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
dkayiwa: Two spots from this thread are still on the read pages: `readAuditLogs.jsp` line 129 shows "No security audit logs found for the given criteria.", and `ReadAuditDetailWebController` line 86 still logs "Error loading security audit detail". Non-blocking, just those two.

api/src/main/java/org/openmrs/module/auditlogweb/api/ReadAuditService.java
sudhanshu-raj: Gd catch, I fixed this and separated the logging audit action into separate component. Keeping this class name bcz we can also store the any other or in future if we gonna save more type of logging/

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/ReadAuditWebController.java
sudhanshu-raj: thanx

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditAOP.java
sudhanshu-raj: Trimmed down to only audit the patient linked entity ignoring the metadata logs for now

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
```diff
@@ -0,0 +1,222 @@
...
+	
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+	
+	private final AppCacheManager appCacheManager;
+	
+	private final AuditLogRecorder auditLogRecorder;
+	
+	@Autowired
+	@Lazy
+	private ReadAuditWorker readAuditWorker;
```
wikumChamith: Can't we do a constructor injection here?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
```diff
@@ -0,0 +1,222 @@
...
+				userUUID = "anonymous";
+				username = "anonymous";
+			}
+			
+			List<ReadAuditEntityMetadata> newTargetEntities = new ArrayList<>();
+			List<String> cacheKeysToSet = new ArrayList<>();
+			List<ReadAuditEntityMetadata> targetEntities = getEntityMetadata(result);
+			for (ReadAuditEntityMetadata targetEntity : targetEntities) {
+				if (targetEntity.getEntityUuid() != null) {
+					String userKey = username != null ? username : (userUUID != null ? userUUID : "anonymous");
```
wikumChamith: Isn't userUUID != null always true here?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
```diff
@@ -0,0 +1,222 @@
...
+				ReadAuditLog readAuditLog = ReadAuditLog.builder().entityName(entityType).eventTime(new Date())
+				        .username(username).userUUID(userUUID).userAgent(userAgent).sessionId(sessionId).ipAddress(ipAddress)
+				        .isReadSuccess(isReadSuccess).build();
+				readAuditLog.setTargets(newTargetEntities);
+				boolean isSubmitted = readAuditWorker.submitTask(readAuditLog);
+				if (isSubmitted) {
+					for (String key : cacheKeysToSet) {
+						appCacheManager.set(key, true);
+					}
+				} else {
```
wikumChamith, line 167: The cache key gets marked as soon as `submitTask` enqueues, but persistence happens later on the worker thread. If both the batch and per-record fallback fail, the record is dropped yet stays deduped for 2 minutes and is never retried. The sync fallback below (158-162) correctly caches only after a successful save, should the async path do the same?

sudhanshu-raj: Right, DB failure can happen and but that log key cache will still be on hold for 2min. If we put the log key in cache after the DB  persist then it will get delayed means suppose if queue has good volume and it takes 10 milliseconds to save the next audit log so here for 10ms for the same key log, we could end up in putting the duplicate audit logs in queue  too . One better option can be let it be get into the queue and if DB persists fail then we will remove that key from the cache hence will allow the new logs to persist.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
sudhanshu-raj: sure

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
```diff
@@ -0,0 +1,220 @@
...
+	
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+	
+	private final AppCacheManager appCacheManager;
+	
+	private final AuditLogRecorder auditLogRecorder;
+	
+	@Lazy
+	private final ReadAuditWorker readAuditWorker;
+
```
wikumChamith: Do we need the lazy annotation now?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
```diff
@@ -0,0 +1,220 @@
...
+	
+	private final AuditLogRecorder auditLogRecorder;
+	
+	@Lazy
+	private final ReadAuditWorker readAuditWorker;
+	
+	public Object auditReadRequest(ProceedingJoinPoint joinPoint) throws Throwable {
+		if (AopUtils.isAopProxy(joinPoint.getTarget())) {
+			return joinPoint.proceed();
+		}
```
wikumChamith, line 62: Do we need this guard now?

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/ReadAuditHelper.java
sudhanshu-raj: We still have the nested proxy, right ? And this helps to only intercept the actual implementation class not the outer proxies, so I guess this guard help us here.

wikumChamith:
image

@sudhanshu-raj it's not clear in the UI what we mean by "target entity UUIDs."

wikumChamith:
Can we also include the patient's name here?

sudhanshu-raj:
> image
> @sudhanshu-raj it's not clear in the UI what we mean by "target entity UUIDs."

I guess we can use a tooltip to describe this, when user will hover the info  icon of this word.

sudhanshu-raj:
> Can we also include the patient's name here?

Here though we not storing the entity name like patient name in the read audit table schema,  as doing for one like Patient will end up with doing for all entity.  But we can make the exception for the Patient case, by checking this separately and then add will add the patient name to the model view. What's you thought  on this ?

wikumChamith:
@sudhanshu-raj, for all the entities, can we also display the patient's name under the primary information?

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,148 @@
...
+	}
+	
+	@PreDestroy
+	public void destroy() {
+		log.info("Stopping ReadAuditWorker background thread");
+		running = false;
+		if (workerThread != null) {
+			workerThread.interrupt();
+		}
+	}
```
wikumChamith, line 77: Won't this abandon un-persisted events? Every restart under load is going to leave an invisible hole in the audit trail.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,148 @@
...
+				saveBatch(batch);
+			}
+			catch (InterruptedException e) {
+				log.debug("ReadAuditWorker thread interrupted, shutting down");
+				Thread.currentThread().interrupt();
+				break;
+			}
+			catch (Exception e) {
+				log.error("Error in ReadAuditWorker execution loop", e);
+			}
```
wikumChamith, line 115: The loop only catches `Exception`. An `Error` (OOM, or `LinkageError`/`NoClassDefFoundError` on redeploy) would kill the worker, while `running` stays true, events keep queueing, but nothing drains them, so auditing silently stops with no signal.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,148 @@
...
+		
+		boolean isBatchLogsSaved = false;
+		try {
+			Context.openSession();
+			auditLogRecorder.logReadAudits(batch);
+			isBatchLogsSaved = true;
+		}
+		catch (Exception e) {
+			log.warn("Failed to save read audit logs in batch, falling back to one-by-one save", e);
+		}
```
wikumChamith: https://github.com/openmrs/openmrs-module-auditlogweb/pull/52/changes#r3612489139

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditLog.java
```diff
@@ -0,0 +1,98 @@
...
+	private String userUUID;
+	
+	@Column(name = "event_time", nullable = false)
+	private Date eventTime;
+	
+	@Column(name = "ip_address", length = 50)
+	private String ipAddress;
+	
+	@Column(name = "user_agent", length = 500)
+	private String userAgent;
```
wikumChamith, line 67: `user_agent` values aren't truncated to the `varchar(500)` column, so a long header would fail the INSERT and the record would be lost permanently, since both the batch and one-by-one fallback would fail. The same risk applies if `entityName` is null going into the NOT-NULL column.

sudhanshu-raj:
> @sudhanshu-raj, for all the entities, can we also display the patient's name under the primary information?

Yeah, we can definitely add this! Just a note: a single audit log event won't always map to a single patient name. For container entities like Cohort (which contain multiple member patients) or when service queries for Patient, Encounter, Obs, etc. return a list of results, multiple patients can be involved in a single read event. So instead of a single patient name under Primary Information, we'll display the list of patient names (or map patient names per target entity). Let me know if that sounds good!

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: yeah, that's harm for storing things into in-memory. Best option i can see is to store the remaining logs to the disk in the .json file during the shutdown, that would be much fast compared to DB store and then later during startup we can store those logs .

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: Fixed, ! Stopping the background loop and the queue to accept new logs in case of any error.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditLog.java
sudhanshu-raj: Added the safety check for this.

sudhanshu-raj:
> > @sudhanshu-raj, for all the entities, can we also display the patient's name under the primary information?
> 
> Yeah, we can definitely add this! Just a note: a single audit log event won't always map to a single patient name. For container entities like Cohort (which contain multiple member patients) or when service queries for Patient, Encounter, Obs, etc. return a list of results, multiple patients can be involved in a single read event. So instead of a single patient name under Primary Information, we'll display the list of patient names (or map patient names per target entity). Let me know if that sounds good!

I have added the patient names to the primary information , but there is one thing for even one Cohort , it can contains thousands of patient which I don't think is efficient to put in the primary information table . So I just skipped it for now , looking for yours thoughts  @wikumChamith .

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
wikumChamith: I don't like the JSON file idea. Do we have an alternative? How about something like a bounded-timeout drain to DB?

sudhanshu-raj: yeah another option can be to drain into DB, though we can improve the  time by saving in one batch .

wikumChamith: Update the PR with a fix then we can decide :)

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,191 @@
...
+	}
+	
+	@PreDestroy
+	public void destroy() {
+		log.info("Stopping ReadAuditWorker background thread");
+		running = false;
+		if (workerThread != null) {
+			workerThread.interrupt();
+		}
+		flushRemainingQueueLogsToDB();
```
wikumChamith, line 76: Can we join(timeout) the worker first, so only one thread drains?

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,191 @@
...
+			}
+		}
+		catch (Exception e) {
+			log.error("Failed to remove read audit keys", e);
+		}
+	}
+	
+	private void flushRemainingQueueLogsToDB() {
+		long startTime = System.currentTimeMillis();
+		long timeoutMs = TIMEOUT_SECONDS * 1000;
```
wikumChamith: The 60s timeout is only checked between batches, so a single `saveBatch` hanging on a slow DB during shutdown could block `destroy()` well past 60s. Also, 60s seems long to hold up a redeploy, could we drop it to ~5–10s?

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,191 @@
...
+	
+	private final AppCacheManager appCacheManager;
+	
+	private final BlockingQueue<ReadAuditLog> queue = new LinkedBlockingQueue<>(10000);
+	
+	private Thread workerThread;
+	
+	private volatile boolean running = true;
+	
+	private final long TIMEOUT_SECONDS = 60;
```
wikumChamith: Should this be static final ?

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
```diff
@@ -0,0 +1,191 @@
...
+		long timeoutMs = TIMEOUT_SECONDS * 1000;
+		
+		log.info("Draining remaining read audit logs to DB. Max timeout: {} seconds", TIMEOUT_SECONDS);
+		
+		int savedCount = 0;
+		while (!queue.isEmpty()) {
+			if (System.currentTimeMillis() - startTime > timeoutMs) {
+				log.warn("Shutdown timeout reached! Aborting and discarding remaining {} read audit logs.", queue.size());
+				break;
+			}
```
wikumChamith: "Successfully flushed {} logs" prints even when 0 were flushed or the timeout discarded some. I think it's worth logging success only when the queue is actually emptied.

wikumChamith (commented):
This is much better. Few minor nits

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: ok done

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: Optimized  to force it to exit and also giving extra sec to let it commit before abort.

api/src/main/java/org/openmrs/module/auditlogweb/ReadAuditWorker.java
sudhanshu-raj: It good find to prevent the race condition , fixed it

ManojLL (approved):

PR #54, opened 2026-07-02
AUDIT-28: Backfill existing data into audit tables

druchniewicz:
Ticket: https://openmrs.atlassian.net/browse/AUDIT-28

A lot of data were empty or unable to read in audit log tables because of missing data in audit tables. For example if you created an encounter only encounterDatetime or dateCreated values were available (basic primitive fields). Any other data that pointed to other tables like Patient/Location/Form/Encounter type etc. were not available because of missing data in their's corresponding audit tables.

We introduced backfill data mechanism that is executed once when Hibernate envers is enabled. Changes made:
* Added AuditBackfillService: a one-time, flag-gated backfill that populates every audit table with existing data under a single baseline revision, run at module start (only when Envers is enabled)
* Registered two global properties: auditlogweb.backfillExistingData.enabled and .completed
* AuditServiceImpl.getRelatedEntitiesInRevision: skip the baseline backfill revision so it doesn't list the entire dataset as "related entities changed in this revision"
* Loaded audited classes via OpenmrsClassLoader (UtilClass.loadClass) so entities from other modules are discoverable, with a Class.forName fallback for unit-test contexts


druchniewicz:
@wikumChamith @dkayiwa can you take a look at PR?

copyOmod.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
wikumChamith: Do we need this?

omod/src/main/resources/config.xml
```diff
@@ -48,6 +48,25 @@
 		</privilege>
 	</privileges>
 
+	<globalProperty>
+		<property>auditlogweb.backfillExistingData.enabled</property>
+		<defaultValue>true</defaultValue>
```
wikumChamith: I think this should be `false` by default. Otherwise every deployment that upgrades will, on first startup,     snapshot every audited table (obs/encounter can be millions of rows)  synchronously on the started() thread.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,337 @@
...
+		revision.setTimestamp(System.currentTimeMillis());
+		revision.setChangedOn(new Date());
+		session.save(revision);
+		session.flush();
+		return revision.getId();
+	}
+	
+	private void runBackfill(Connection connection, List<TableMapping> mappings, int revId) throws SQLException {
+		try (Statement off = connection.createStatement()) {
+			off.execute("SET FOREIGN_KEY_CHECKS = 0");
```
wikumChamith: Doesn't this only work on MySQL?

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,337 @@
...
+				tx.commit();
+			}
+			catch (RuntimeException e) {
+				tx.rollback();
+				throw e;
+			}
+		}
+		
+		administrationService.setGlobalProperty(GP_BACKFILL_REVISION, String.valueOf(revId));
+		administrationService.setGlobalProperty(GP_BACKFILL_COMPLETED, "true");
```
wikumChamith: Won't this set `completed=true` even if the backfill partially fails?

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Properties;
+import java.util.Set;
+import java.util.TreeMap;
+
+@Component("auditlogweb.auditBackfillService")
+@RequiredArgsConstructor
+public class AuditBackfillService {
```
dkayiwa, line 30: Can we get some tests on this before it goes in? The class ships untested, and the new baseline-skip in AuditServiceImpl.getRelatedEntitiesInRevision has no test either (the only test changes in the PR are stubs that keep existing tests green). The part that worries me is the gating: if the enabled/completed checks regress, a deployment that opted out gets its whole database snapshotted into the audit tables on the next restart, and nothing in CI would notice. Similarly for isBaselineRevision: if that comparison breaks, the related-entities view either lists the entire dataset for the baseline revision (exactly what the skip is meant to prevent) or quietly returns empty for real revisions. The gating short-circuits, reuseRevisionId and isBaselineRevision are all coverable with the mockStatic(Context) pattern the existing service tests already use.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,337 @@
...
+				}
+				final int revision = revId;
+				session.doWork(connection -> runBackfill(connection, mappings, revision));
+				tx.commit();
+			}
+			catch (RuntimeException e) {
+				tx.rollback();
+				throw e;
+			}
+		}
```
dkayiwa: I'd rethink running the whole backfill in one transaction. On the databases where this feature matters most (years of obs and encounter rows) this is one enormous transaction: hours of inserts, a lot of undo and lock pressure, and an interruption near the end rolls back everything, including the tables that already succeeded. The NOT EXISTS guard already makes re-runs idempotent, so committing per table would bound the damage and let a crashed run resume where it stopped. If you do that, also write GP_BACKFILL_REVISION right after the revision row commits: today it is only saved at the very end, so a run that dies midway would mint a second baseline revision on retry and reuseRevisionId never gets a chance to help. Not something I'll hold the PR over on its own, but it is much easier to shape now than after it has run on someone's production database.

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -350,6 +352,15 @@ public long countRevisionsAcrossEntitiesWithEntityType(Integer userId, Date star
 	
 	@Override
 	public List<AuditEntity<?>> getRelatedEntitiesInRevision(Class<?> entityClass, Object entityId, int revisionId) {
+		// The one-time backfill assigns every pre-existing row to a single "baseline" revision.
+		// That revision is a bulk import, not a real transaction, so treating it as "these entities changed
+		// together" would list the entire audited dataset. Skip it for this feature.
+		AuditBackfillService backfillService = Context.getRegisteredComponent("auditlogweb.auditBackfillService",
+		    AuditBackfillService.class);
+		if (backfillService != null && backfillService.isBaselineRevision(revisionId)) {
```
dkayiwa: Context.getRegisteredComponent never returns null; when the bean is missing it throws APIException (ServiceContext wraps the BeansException), so this null check is dead code and the graceful fallback it suggests doesn't exist. Since AuditBackfillService lives in the same Spring context as this service, I'd inject it instead: a final field plus a constructor-arg next to the dao one in moduleApplicationContext.xml. That also avoids repeating the bean lookup on every call.

scripts/audit_test_helpers.sql
```diff
@@ -0,0 +1,148 @@
+-- ============================================================================
```
dkayiwa: Same question as Wikum's on copyOmod.sh, for this file: is it meant to ship in the repo, or was it part of local testing? If it stays, a pointer from the README would help people find it.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
dkayiwa: It's MySQL-only in more places than this statement: quoteIdentifier() uses backticks, so swapping just the FOREIGN_KEY_CHECKS calls won't make it portable. Worth knowing the current failure mode too: on any other database the first SET statement throws, the whole backfill aborts, completed never gets set, and the module re-attempts and error-logs the backfill on every restart. As far as I can tell the FK toggling isn't needed at all, though. The only FK constraints on the audit tables are the REV references to revision_entity, and the baseline revision row is inserted in the same transaction as the copies, so the checks would pass anyway. Dropping the two SET statements and quoting identifiers with DatabaseMetaData.getIdentifierQuoteString() would get this most of the way to portable.

copyOmod.sh
druchniewicz: no, it was my fault. removed

scripts/audit_test_helpers.sql
druchniewicz: no, it was my fault. removed

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: tests added

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: corrected, now it is set to true only if full success

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: corrected

omod/src/main/resources/config.xml
druchniewicz: corrected.
it was initially set to true, because you need to enable hibernate envers first to run this backfilling. If you enable hibernate envers then this backfilling would be executed automatically.

druchniewicz:
@wikumChamith @dkayiwa PR updated

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
```diff
@@ -391,9 +402,7 @@ private Object fetchPreviousRevision(AuditEntity<?> entity, Object currentEntity
 				//Here if this condition becomes true then most probably, this is id or uuid
 				entityId = Integer.parseInt(entityId.toString());
 			}
-			catch (NumberFormatException e) {
-				//If this exception occurred then it may be the uuid, which we're trying to convert to int, so leave it as string
-			}
+			catch (NumberFormatException ignored) {}
```
wikumChamith: I know this was empty before too, but wouldn't it be better if we added a debug-level log here rather than having an empty catch clause?

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,417 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.TreeMap;
+
+@Component("auditlogweb.auditBackfillService")
+@RequiredArgsConstructor
+public class AuditBackfillService {
```
wikumChamith, line 30: TBH,  it would be better if we extracted a separate DAO class for database interaction here. This would also let us write tests for important logic like `backfillTable`'s SQL generation, `resolveAuditedTableMappings`, and the new `orderByAuditTableDependencies` topological sort, which currently have no coverage.

Even if we don't extract a DAO class, we need test coverage for those methods :)

api/src/main/java/org/openmrs/module/auditlogweb/api/impl/AuditServiceImpl.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: corrected

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -0,0 +1,140 @@
...
+	
+	private static final Logger log = LoggerFactory.getLogger(AuditBackfillService.class);
+	
+	public static final String GP_BACKFILL_ENABLED = "auditlogweb.backfillExistingData.enabled";
+	
+	public static final String GP_BACKFILL_COMPLETED = "auditlogweb.backfillExistingData.completed";
+	
+	public static final String GP_BACKFILL_REVISION = "auditlogweb.backfillExistingData.revision";
```
wikumChamith, line 38: Don't we need to declare this in the `config.xml`?

wikumChamith (approved):
One minor nit, otherwise LGTM!

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
druchniewicz: I've added it to keep it consistent

PR #55, opened 2026-07-03
Return display values and ISO-8601 dates in audit field diffs

RajPrakash681:

Makes the audit field diffs human-readable instead of exposing raw backend representations.

- `AuditFieldDiff` now carries nullable `oldDisplay` / `currentDisplay` alongside the raw `oldValue` / `currentValue` (raw values retained for auditability).
- Entity-reference values — previously serialized as `SimpleClassName#id` (e.g. `Concept#88`) — are resolved to a human label looked up live by id via the OpenMRS services (Concept, User, Location, Person), e.g. `Malaria (Concept#88)`. Resolution does not traverse the Envers snapshot's lazy associations and degrades to `null` when the id can't be resolved.
- `java.util.Date` diff values are serialized as ISO-8601 (`Instant.ofEpochMilli(getTime())`, safe for `java.sql.Date`) instead of Java `Date.toString()`.
- The new display fields are copied through `AuditServiceImpl.extractChangedFields`, so the `/auditlogs/patients` endpoint retains them.

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-59

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/AuditFieldDiff.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/AuditFieldDiff.java
@@ -25,4 +25,8 @@ public class AuditFieldDiff {
 	private String currentValue;
 	
 	private boolean changed;
+	
+	private String oldDisplay;
+	
+	private String currentDisplay;
 }
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
@@ -13,6 +13,11 @@ import org.apache.commons.lang3.StringUtils;
 import org.hibernate.envers.Audited;
 import org.hibernate.proxy.HibernateProxy;
 import org.openmrs.BaseOpenmrsObject;
+import org.openmrs.Concept;
+import org.openmrs.Location;
+import org.openmrs.Person;
+import org.openmrs.User;
+import org.openmrs.api.context.Context;
 import org.openmrs.util.OpenmrsClassLoader;
 import org.openmrs.module.auditlogweb.api.dto.AuditFieldDiff;
 import org.reflections.Reflections;
@@ -28,6 +33,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.ZoneId;
@@ -192,9 +198,11 @@ public class UtilClass {
 			String currVal;
 			boolean failedOld = false;
 			boolean failedCurr = false;
+			Object currFieldValue = null;
+			Object oldFieldValue = null;
 			
 			try {
-				Object currFieldValue = field.get(currentEntity);
+				currFieldValue = field.get(currentEntity);
 				currVal = serializeFieldValue(currFieldValue);
 			}
 			catch (Exception e) {
@@ -204,7 +212,7 @@ public class UtilClass {
 			}
 			
 			try {
-				Object oldFieldValue = oldEntity != null ? field.get(oldEntity) : null;
+				oldFieldValue = oldEntity != null ? field.get(oldEntity) : null;
 				oldVal = serializeFieldValue(oldFieldValue);
 			}
 			catch (Exception e) {
@@ -219,7 +227,16 @@ public class UtilClass {
 			}
 			
 			boolean isDifferent = !Objects.equals(oldVal, currVal);
-			diffs.add(new AuditFieldDiff(field.getName(), oldVal, currVal, isDifferent));
+			AuditFieldDiff diff = new AuditFieldDiff();
+			diff.setFieldName(field.getName());
+			diff.setOldValue(oldVal);
+			diff.setCurrentValue(currVal);
+			diff.setChanged(isDifferent);
+			if (isDifferent) {
+				diff.setOldDisplay(resolveDisplayValue(oldFieldValue));
+				diff.setCurrentDisplay(resolveDisplayValue(currFieldValue));
+			}
+			diffs.add(diff);
 		}
 		return diffs;
 	}
@@ -348,6 +365,10 @@ public class UtilClass {
 			return "";
 		}
 		
+		if (value instanceof Date) {
+			return Instant.ofEpochMilli(((Date) value).getTime()).toString();
+		}
+		
 		String actualClassName = getActualClassName(value);
 		if (isPrimitiveOrWrapper(value.getClass())) {
 			return String.valueOf(value);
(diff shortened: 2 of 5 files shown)


RajPrakash681:
@ManojLL please have a look.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -432,6 +453,40 @@ private static String serializeGenericObject(Object value, String actualClassNam
...
+				name = concept != null ? concept.getDisplayString() : null;
+			} else if (value instanceof User) {
+				User user = Context.getUserService().getUser(id);
+				name = user != null ? user.getDisplayString() : null;
+			} else if (value instanceof Location) {
+				Location location = Context.getLocationService().getLocation(id);
+				name = location != null ? location.getName() : null;
+			} else if (value instanceof Person) {
+				Person person = Context.getPersonService().getPerson(id);
+				name = (person != null && person.getPersonName() != null) ? person.getPersonName().getFullName() : null;
```
wikumChamith: Only a few entity types are resolved here. A lot of important entity types are going to fall back to the raw `Type#id`. Since most OpenMRS metadata implement `OpenmrsMetadata.getName()`, we can cover a large swath in one branch:

```java
else if (value instanceof OpenmrsMetadata) {
    name = ((OpenmrsMetadata) value).getName();
}
```

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -432,6 +453,40 @@ private static String serializeGenericObject(Object value, String actualClassNam
...
+			return null;
+		}
+		Integer id = getIdFromObject(value);
+		if (id == null) {
+			return null;
+		}
+		try {
+			String name = null;
+			if (value instanceof Concept) {
+				Concept concept = Context.getConceptService().getConcept(id);
```
wikumChamith: This resolves against the live entity, so `oldDisplay` on an old revision shows the entity's name today, not at the time of the revision.

RajPrakash681:
Thanks  @wikumChamith  added the OpenmrsMetadata branch, and kept resolution live with the `Type#id `retained so the entity's still traceable. Could you take another look and suggest accordingly 🙂

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/AuditFieldDiff.java
```diff
@@ -25,4 +25,8 @@ public class AuditFieldDiff {
 	private String currentValue;
 	
 	private boolean changed;
+	
+	private String oldDisplay;
+	
+	private String currentDisplay;
```
dkayiwa, line 31: These new fields reach JSON clients fine (`AuditLogDetailDTO` serializes them through the REST controllers), but the server-rendered audit page doesn't use them. `viewAudit.jsp` still prints `${diff.oldValue}` / `${diff.currentValue}`, and `ViewAuditController` already calls `computeFieldDiffs`, so it resolves `oldDisplay`/`currentDisplay` and then discards them. Net effect: on the main "View Audit" page users still see the raw `Concept#88` this PR set out to replace.

Was scoping the human-readable labels to the JSON API intentional? If the HTML view was meant to benefit too, rendering the label with a fallback is almost free here since the value is already computed:

```jsp
<td><c:out value="${not empty diff.oldDisplay ? diff.oldDisplay : diff.oldValue}" /></td>
<td><c:out value="${not empty diff.currentDisplay ? diff.currentDisplay : diff.currentValue}" /></td>
```

Not blocking, just flagging the gap against the stated goal.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -199,7 +208,16 @@ public static List<AuditFieldDiff> computeFieldDiffs(Class<?> clazz, Object oldE
...
 			boolean isDifferent = !Objects.equals(oldVal, currVal);
-			diffs.add(new AuditFieldDiff(field.getName(), oldVal, currVal, isDifferent));
+			AuditFieldDiff diff = new AuditFieldDiff();
+			diff.setFieldName(field.getName());
+			diff.setOldValue(oldVal);
+			diff.setCurrentValue(currVal);
+			diff.setChanged(isDifferent);
+			if (isDifferent) {
+				diff.setOldDisplay(resolveDisplayValue(oldFieldValue));
+				diff.setCurrentDisplay(resolveDisplayValue(currFieldValue));
```
dkayiwa: Each changed reference field resolved here fires a live lookup (`getConcept` / `getUser` / `getLocation` / `getPerson`). That's cheap for a single revision, but `getEntityDetailedAudit` iterates every revision of a patient and `mapAuditEntitiesToDetails` iterates a whole page of entries, and both call into this per changed field. So `/auditlogs/patients` ends up issuing on the order of (revisions x reference-fields) single-row queries per request, with no caching.

Correctness is fine (the lookups are wrapped in try/catch), this is purely about query volume, so not blocking. If it gets noticeable on entities with long histories, memoizing within the request keyed by type + id would collapse the repeats.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -432,6 +454,47 @@ private static String serializeGenericObject(Object value, String actualClassNam
...
+		}
+		try {
+			String name = null;
+			if (value instanceof Concept) {
+				Concept concept = Context.getConceptService().getConcept(id);
+				name = concept != null ? concept.getDisplayString() : null;
+			} else if (value instanceof User) {
+				User user = Context.getUserService().getUser(id);
+				name = user != null ? user.getDisplayString() : null;
+			} else if (value instanceof Location) {
```
wikumChamith: Location implements OpenmrsMetadata, so this branch is now redundan now.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -432,6 +454,47 @@ private static String serializeGenericObject(Object value, String actualClassNam
 		return "";
 	}
 	
+	/**
+	 * Resolves an entity reference to a human label, looked up live by id, as "Name (Type#id)". The
+	 * label reflects the entity's current name; the retained Type#id token identifies the exact
+	 * referenced entity for revision-accurate auditing.
+	 */
```
wikumChamith: The doc says the label is "looked up live by id", but that's only true for the Concept/User/Location/Person branches.

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/AuditFieldDiff.java
RajPrakash681: yes, @dkayiwa you are right.  will validate the approach...

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -328,6 +346,10 @@ public static String serializeFieldValue(Object value) {
 			return "";
 		}
 		
+		if (value instanceof Date) {
+			return Instant.ofEpochMilli(((Date) value).getTime()).toString();
```
wikumChamith: `Instant.ofEpochMilli(...)` renders in UTC. Date-only fields (e.g. `birthdate`) built at local midnight then shift a day in non-UTC zones, displaying the wrong day. Tests miss this since epoch 0 is UTC-invariant. Consider using the system zone/`LocalDate` for date-only fields, and add a non-UTC test.

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/UtilClass.java
```diff
@@ -432,6 +454,47 @@ private static String serializeGenericObject(Object value, String actualClassNam
...
+				User user = Context.getUserService().getUser(id);
+				name = user != null ? user.getDisplayString() : null;
+			} else if (value instanceof Location) {
+				Location location = Context.getLocationService().getLocation(id);
+				name = location != null ? location.getName() : null;
+			} else if (value instanceof Person) {
+				Person person = Context.getPersonService().getPerson(id);
+				name = (person != null && person.getPersonName() != null) ? person.getPersonName().getFullName() : null;
+			} else if (value instanceof OpenmrsMetadata) {
+				name = ((OpenmrsMetadata) value).getName();
```
wikumChamith: This reads `getName()` off the snapshot proxy, unlike the live-lookup branches. If it's a lazy proxy and the session is closed, it'll throw `LazyInitializationException`, which gets swallowed and results in a null display where the other branches would succeed. Can you confirm the session is open at DTO-build time?

wikumChamith:
@RajPrakash681 also  fix the merge conflicts

PR #56, opened 2026-07-05
AUDIT-60 : Add a fallback for audit log context value due to filter chain order change

sudhanshu-raj:
During the authenticaton event  like login success/failed when we do from O3 interface, as this directly hits the REST endpoint to the webservices module. And due to Authorization filter of webservice, it catches the request first and authenticates the user there only and since till now AuditLogContext not got initiliazed, during auditing this authetication event context got null and that's end up in saving empty value for device metadata.

This PR fixes it, changes the order module startup order so that `AuditContextFilter` come before  the `AuthorizationFilter `  and we can get the metadata value before.

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-60

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
@@ -28,11 +28,6 @@ import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
-import org.springframework.web.context.request.RequestContextHolder;
-import org.springframework.web.context.request.ServletRequestAttributes;
-
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpSession;
 
 // It logs the event related to authentication like login success/failed and account locked.
 @Aspect
@@ -58,31 +53,6 @@ public class AuthenticationAdvice {
 			sessionId = ctx.getSessionId();
 			ipAddress = ctx.getIpAddress();
 			userAgent = ctx.getUserAgent();
-		} else {
-			try {
-				ServletRequestAttributes attributes = (ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
-				if (attributes != null) {
-					HttpServletRequest request = attributes.getRequest();
-					if (request != null) {
-						userAgent = request.getHeader("User-Agent");
-						
-						String forwarded = request.getHeader("X-Forwarded-For");
-						if (forwarded != null && !forwarded.isEmpty()) {
-							ipAddress = forwarded.split(",")[0].trim();
-						} else {
-							ipAddress = request.getRemoteAddr();
-						}
-						
-						HttpSession session = request.getSession(false);
-						if (session != null) {
-							sessionId = session.getId();
-						}
-					}
-				}
-			}
-			catch (Exception e) {
-				log.warn("Failed to extract context via RequestContextHolder", e);
-			}
 		}
 		
 		try {
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
@@ -28,6 +28,11 @@ import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
 
 // It logs the event related to authentication like login success/failed and account locked.
 @Aspect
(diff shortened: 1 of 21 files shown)


sudhanshu-raj:
Hi @wikumChamith @ManojLL please review this

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -45,9 +50,40 @@ public class AuthenticationAdvice {
...
+		String ipAddress = null;
+		String userAgent = null;
+		
+		if (ctx != null) {
+			sessionId = ctx.getSessionId();
+			ipAddress = ctx.getIpAddress();
+			userAgent = ctx.getUserAgent();
+		} else {
+			try {
+				ServletRequestAttributes attributes = (ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
```
dkayiwa: How did you confirm that `RequestContextHolder` actually has the request bound at this point? Tracing the filter chain, I don't think it does in the exact case this targets.

Both `AuditContextFilter` (auditlogweb) and `AuthorizationFilter` (webservices) are module filters, so they run inside OpenMRS's `ModuleFilter`, which sits in the servlet filter chain ahead of the `DispatcherServlet`. `RequestContextHolder` gets bound by the `DispatcherServlet` (`FrameworkServlet.processRequest`), and openmrs-core's `web.xml` registers no `RequestContextListener` or `RequestContextFilter`, so nothing binds the request to the thread during the filter phase. When `AuthorizationFilter` authenticates a REST login, we're still in that filter phase and `AuditContextFilter` hasn't run yet, which is exactly why `ctx` is null here. But `RequestContextHolder.getRequestAttributes()` would be null for the same reason, so this branch falls straight through and the audit event ends up with empty IP / User-Agent / session again, the original bug.

The new unit test doesn't catch this because it calls `RequestContextHolder.setRequestAttributes(...)` by hand, simulating a binding the servlet container hasn't actually performed at filter time.

If I've got the ordering wrong, please point me at where the request gets bound before this runs. Otherwise the real fix is likely to make `AuditContextFilter` run ahead of the webservices `AuthorizationFilter`, rather than reading from `RequestContextHolder`.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -45,9 +50,40 @@ public class AuthenticationAdvice {
...
+				if (attributes != null) {
+					HttpServletRequest request = attributes.getRequest();
+					if (request != null) {
+						userAgent = request.getHeader("User-Agent");
+						
+						String forwarded = request.getHeader("X-Forwarded-For");
+						if (forwarded != null && !forwarded.isEmpty()) {
+							ipAddress = forwarded.split(",")[0].trim();
+						} else {
+							ipAddress = request.getRemoteAddr();
```
dkayiwa: If this fallback stays, the IP resolution here (`X-Forwarded-For` then `getRemoteAddr`), plus the User-Agent and session extraction just above, duplicate `AuditContextFilter.resolveClientIp` / `buildContext` almost verbatim. Worth pulling into one shared helper so the two paths can't drift apart later. Non-blocking.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
sudhanshu-raj: Yeah you are right, `RequestContextHolder`  will get set after the request hits `DispatcherServlet `  and since all filters run before `DispatcherServlet` including the `AuthorizationFilter`  so the time when we will audit the authentication event `RequestContextHolder` will have probably null and thus that fallback is not useful.

Actually when I was testing on ref app, after I loaded the new build audit module, OpenMRS set the    `AuditContextFilter` at top bcz in the openmrs-core we do set the filter at the top for new added module and thus for that phase  of testing `AuditContextFilter` is in the first order of filter chain that's why it worked and I thought new change has worked 😅 but no, I should have restarted the whole backend.

And yeah seems like we need to let `AuditContextFilter` come before `AuthorizationFilter` filter chain, so that's why i loading the audit module after webservice module so that it's filter get at before in chain and we can fill the metadata first.

wikumChamith:
@sudhanshu-raj, have you tested the filter ordering on a cold restart?

sudhanshu-raj:
> @sudhanshu-raj, have you tested the filter ordering on a cold restart?

Yeah I did ordering test, and it's coming before the Authorization filter.

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
```diff
@@ -45,9 +45,15 @@ public class AuthenticationAdvice {
...
-		String userAgent = ctx != null ? ctx.getUserAgent() : null;
+		String sessionId = null;
+		String ipAddress = null;
+		String userAgent = null;
+		
+		if (ctx != null) {
+			sessionId = ctx.getSessionId();
+			ipAddress = ctx.getIpAddress();
+			userAgent = ctx.getUserAgent();
+		}
```
wikumChamith, line 56: Do we need this refactor?

wikumChamith (approved):
Something minor, otherwise LGTM!

api/src/main/java/org/openmrs/module/auditlogweb/api/aop/AuthenticationAdvice.java
sudhanshu-raj: Looks cleaner, we can keep it.

PR #58, opened 2026-07-27
AUDIT-61: Create the REST APIs for the Security Audit Logging

sudhanshu-raj:
Created the REST APIs to fetch the security audit logs.

The main API for fetching the logs based on different params, like  : 
`openmrs/ws/rest/v1/securityauditlogs
`
`openmrs/ws/rest/v1/securityauditlogs?logId=21
`
`openmrs/ws/rest/v1/securityauditlogs?eventType=SESSION_TIMEOUT&startDate=22/07/2026
`

And the second one for fetching the related logs reference to that session id, like :
`openmrs/ws/rest/v1/securityauditlogs/releatedAudits?sessionId=5048796D4BBFA09EAB8DA6D565B40E51
`

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-61

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
@@ -334,14 +334,20 @@ public interface AuditService {
 	 * Retrieves the most recent N security events from the same session (for related activity).
 	 *
 	 * @param sessionId the session ID to filter by
-	 * @param page defines the particular page in a paginated view
 	 * @param size defines how many logs want in one go
+	 * @param page defines the particular page in a paginated view
 	 * @return a list of {@link AuditSecurityEvent} ordered by eventTime descending
 	 */
 	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
 	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int size, int page);
 	
-	@Authorized(AuditLogConstants.VIEW_READ_AUDIT_LOGS)
+	/**
+	 * Counts the related security audit events with session ID filter.
+	 *
+	 * @param sessionId the session ID to filter by
+	 * @return the number of logs found
+	 */
+	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
 	long countRelatedSecurityEvents(String sessionId);
 	
 }
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
@@ -334,12 +334,12 @@ public interface AuditService {
 	 * Retrieves the most recent N security events from the same session (for related activity).
 	 *
 	 * @param sessionId the session ID to filter by
-	 * @param size defines how many logs want in one go
 	 * @param page defines the particular page in a paginated view
+	 * @param size defines how many logs want in one go
 	 * @return a list of {@link AuditSecurityEvent} ordered by eventTime descending
 	 */
 	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
-	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int size, int page);
+	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int page, int size);
 	
 	/**
 	 * Counts the related security audit events with session ID filter.
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
@@ -334,10 +334,14 @@ public interface AuditService {
 	 * Retrieves the most recent N security events from the same session (for related activity).
 	 *
 	 * @param sessionId the session ID to filter by
-	 * @param limit the maximum number of events to return
+	 * @param page defines the particular page in a paginated view
+	 * @param size defines how many logs want in one go
 	 * @return a list of {@link AuditSecurityEvent} ordered by eventTime descending
 	 */
 	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
-	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int limit);
+	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int size, int page);
+	
+	@Authorized(AuditLogConstants.VIEW_READ_AUDIT_LOGS)
+	long countRelatedSecurityEvents(String sessionId);
 	
 }
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
@@ -9,6 +9,7 @@
  */
 package org.openmrs.module.auditlogweb.api.dto;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
 import lombok.Builder;
 import lombok.Data;
 import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
@@ -27,6 +28,7 @@ public class SecurityAuditLogDTO {
 	
 	private String userUuid;
 	
+	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd/MM/yyyy HH:mm:ss", timezone = "GMT")
 	private Date eventTime;
 	
 	private String ipAddress;
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
@@ -1,0 +1,40 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.auditlogweb.api.dto;
+
+import lombok.Builder;
+import lombok.Data;
+import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
+
+import java.util.Date;
+
+@Data
+@Builder
+public class SecurityAuditLogDTO {
+	
+	private Integer id;
+	
+	private AuditSecurityEventType eventType;
+	
+	private String username;
+	
+	private String userUuid;
+	
+	private Date eventTime;
+	
+	private String ipAddress;
+	
+	private String userAgent;
+	
+	private String sessionId;
+	
+	private String details;
+	
+}
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityLogResponseDTO.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityLogResponseDTO.java
@@ -1,0 +1,31 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.auditlogweb.api.dto;
+
+import lombok.Builder;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@Builder
+public class SecurityLogResponseDTO {
+	
+	private long totalLogs;
+	
+	private int currentLogs;
+	
+	private int totalPages;
+	
+	private int currentPage;
+	
+	private List<SecurityAuditLogDTO> securityAuditLogs;
+	
+}
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
@@ -17,4 +17,6 @@ public final class AuditLogConstants {
 	public static final String VIEW_SECURITY_AUDIT_LOGS = "View Security Audit Logs";
 	
 	public static final String VIEW_READ_AUDIT_LOGS = "View Read Audit Logs";
+	
+	public static final Integer MAX_PAGE_SIZE = 200;
 }
--- a/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
+++ b/api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
@@ -18,5 +18,5 @@ public final class AuditLogConstants {
 	
 	public static final String VIEW_READ_AUDIT_LOGS = "View Read Audit Logs";
 	
-	public static final Integer MAX_PAGE_SIZE = 200;
+	public static final int MAX_PAGE_SIZE = 200;
 }
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -13,6 +13,7 @@ import lombok.RequiredArgsConstructor;
 import org.openmrs.module.auditlogweb.ReadAuditLog;
 import org.openmrs.module.auditlogweb.api.ReadAuditService;
 import org.openmrs.module.auditlogweb.api.dto.ReadAuditLogResponseDTO;
+import org.openmrs.module.auditlogweb.api.utils.AuditLogConstants;
 import org.openmrs.module.auditlogweb.api.utils.UtilClass;
 import org.openmrs.module.auditlogweb.api.dto.ReadAuditLogDTO;
 import org.openmrs.module.webservices.rest.web.RestConstants;
@@ -65,6 +66,7 @@ public class ReadAuditRestController {
 		if (size <= 0) {
 			size = 15;
 		}
+		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
 		
 		Date start = UtilClass.parseDate(startDate, false);
 		Date end = UtilClass.parseDate(endDate, true);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -57,7 +57,7 @@ public class ReadAuditRestController {
 			List<ReadAuditLogDTO> readAuditLogsDTO = readAuditService
 			        .mapToReadAuditLogDTO(Collections.singletonList(readAuditLog));
 			return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).readAuditLogs(readAuditLogsDTO)
-			        .totalPages(1).currentPage(0).build();
+			        .totalPages(1).build();
 		}
 		
 		if (page < 0) {
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -60,13 +60,8 @@ public class ReadAuditRestController {
 			        .totalPages(1).build();
 		}
 		
-		if (page < 0) {
-			page = 0;
-		}
-		if (size <= 0) {
-			size = 15;
-		}
-		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
+		size = UtilClass.sanitizePageSizeValue(size);
+		page = UtilClass.sanitizePageValue(page, size);
 		
 		Date start = UtilClass.parseDate(startDate, false);
 		Date end = UtilClass.parseDate(endDate, true);
@@ -92,13 +87,8 @@ public class ReadAuditRestController {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
 		}
 		
-		if (page < 0) {
-			page = 0;
-		}
-		if (size <= 0) {
-			size = 15;
-		}
-		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
+		size = UtilClass.sanitizePageSizeValue(size);
+		page = UtilClass.sanitizePageValue(page, size);
 		
 		List<ReadAuditLog> relatedAudits = readAuditService.getRelatedReadLogs(sessionId, page, size);
 		long totalCount = readAuditService.countRelatedReadLogs(sessionId);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -88,7 +88,7 @@ public class ReadAuditRestController {
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
 		
-		if (sessionId == null || sessionId.isEmpty()) {
+		if (sessionId == null || sessionId.trim().isEmpty()) {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
 		}
 		
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -98,6 +98,7 @@ public class ReadAuditRestController {
 		if (size <= 0) {
 			size = 15;
 		}
+		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
 		
 		List<ReadAuditLog> relatedAudits = readAuditService.getRelatedReadLogs(sessionId, page, size);
 		long totalCount = readAuditService.countRelatedReadLogs(sessionId);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -81,7 +81,7 @@ public class ReadAuditRestController {
 		
 	}
 	
-	@GetMapping("/releatedAudits")
+	@GetMapping("/relatedAudits")
 	public ReadAuditLogResponseDTO fetchRelatedAudits(@RequestParam(value = "sessionId") String sessionId,
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -55,7 +55,8 @@ public class ReadAuditRestController {
 			}
 			List<ReadAuditLogDTO> readAuditLogsDTO = readAuditService
 			        .mapToReadAuditLogDTO(Collections.singletonList(readAuditLog));
-			return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).readAuditLogs(readAuditLogsDTO).build();
+			return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).readAuditLogs(readAuditLogsDTO)
+			        .totalPages(1).currentPage(0).build();
 		}
 		
 		if (page < 0) {
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
@@ -51,8 +51,7 @@ public class ReadAuditRestController {
 		if (logId != null) {
 			ReadAuditLog readAuditLog = readAuditService.getReadAuditLogById(logId);
 			if (readAuditLog == null) {
-				return ReadAuditLogResponseDTO.builder().totalLogs(0).currentLogs(0).readAuditLogs(Collections.emptyList())
-				        .build();
+				throw new ResponseStatusException(HttpStatus.NOT_FOUND, "No log found for this logId");
 			}
 			List<ReadAuditLogDTO> readAuditLogsDTO = readAuditService
 			        .mapToReadAuditLogDTO(Collections.singletonList(readAuditLog));
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -67,13 +67,8 @@ public class SecurityAuditRestController {
 			        .totalPages(1).build();
 		}
 		
-		if (page < 0) {
-			page = 0;
-		}
-		if (size <= 0) {
-			size = 15;
-		}
-		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
+		size = UtilClass.sanitizePageSizeValue(size);
+		page = UtilClass.sanitizePageValue(page, size);
 		
 		Date start = UtilClass.parseDate(startDate, false);
 		Date end = UtilClass.parseDate(endDate, true);
@@ -98,13 +93,8 @@ public class SecurityAuditRestController {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
 		}
 		
-		if (page < 0) {
-			page = 0;
-		}
-		if (size <= 0) {
-			size = 15;
-		}
-		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
+		size = UtilClass.sanitizePageSizeValue(size);
+		page = UtilClass.sanitizePageValue(page, size);
 		
 		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, page, size);
 		long totalCount = auditService.countRelatedSecurityEvents(sessionId);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -14,6 +14,7 @@ import org.openmrs.module.auditlogweb.AuditSecurityEvent;
 import org.openmrs.module.auditlogweb.api.AuditService;
 import org.openmrs.module.auditlogweb.api.dto.SecurityAuditLogDTO;
 import org.openmrs.module.auditlogweb.api.dto.SecurityLogResponseDTO;
+import org.openmrs.module.auditlogweb.api.utils.AuditLogConstants;
 import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
 import org.openmrs.module.auditlogweb.api.utils.UtilClass;
 import org.openmrs.module.webservices.rest.web.RestConstants;
@@ -72,6 +73,7 @@ public class SecurityAuditRestController {
 		if (size <= 0) {
 			size = 15;
 		}
+		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
 		
 		Date start = UtilClass.parseDate(startDate, false);
 		Date end = UtilClass.parseDate(endDate, true);
@@ -92,7 +94,7 @@ public class SecurityAuditRestController {
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
 		
-		if (sessionId == null || sessionId.isEmpty()) {
+		if (sessionId == null || sessionId.trim().isEmpty()) {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
 		}
 		
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -104,7 +104,7 @@ public class SecurityAuditRestController {
 			size = 15;
 		}
 		
-		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, size, page);
+		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, page, size);
 		long totalCount = auditService.countRelatedSecurityEvents(sessionId);
 		int totalPages = UtilClass.computeTotalPages(totalCount, size);
 		
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -49,7 +49,7 @@ public class SecurityAuditRestController {
 			throw new IllegalArgumentException("Please provide a valid log ID");
 		}
 		
-		if (eventType != null) {
+		if (eventType != null && !eventType.trim().isEmpty()) {
 			AuditSecurityEventType parsed = AuditSecurityEventType.fromName(eventType);
 			if (parsed == null || parsed == AuditSecurityEventType.UNKNOWN) {
 				throw new IllegalArgumentException("Invalid eventType " + eventType);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -104,6 +104,7 @@ public class SecurityAuditRestController {
 		if (size <= 0) {
 			size = 15;
 		}
+		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
 		
 		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, page, size);
 		long totalCount = auditService.countRelatedSecurityEvents(sessionId);
--- a/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
+++ b/omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
@@ -18,7 +18,6 @@ import org.openmrs.module.auditlogweb.api.utils.AuditSecurityEventType;
 import org.openmrs.module.auditlogweb.api.utils.UtilClass;
 import org.openmrs.module.webservices.rest.web.RestConstants;
 import org.springframework.http.HttpStatus;
-import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
@@ -34,9 +33,9 @@ import java.util.List;
 @RequestMapping("/rest/" + RestConstants.VERSION_1 + "/securityauditlogs")
 @RequiredArgsConstructor
 public class SecurityAuditRestController {
-	
+
 	private final AuditService auditService;
-	
+
 	@GetMapping
 	public SecurityLogResponseDTO fetchSecurityAudits(@RequestParam(value = "logId", required = false) Integer logId,
 	        @RequestParam(value = "eventType", required = false) String eventType,
@@ -45,18 +44,18 @@ public class SecurityAuditRestController {
 	        @RequestParam(value = "endDate", required = false) String endDate,
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
-		
+
 		if (logId != null && logId <= 0) {
 			throw new IllegalArgumentException("Please provide a valid log ID");
 		}
-		
+
 		if (eventType != null && !eventType.trim().isEmpty()) {
 			AuditSecurityEventType parsed = AuditSecurityEventType.fromName(eventType);
 			if (parsed == null || parsed == AuditSecurityEventType.UNKNOWN) {
 				throw new IllegalArgumentException("Invalid eventType " + eventType);
 			}
 		}
-		
+
 		if (logId != null) {
 			AuditSecurityEvent securityEvent = auditService.getSecurityEventById(logId);
 			if (securityEvent == null) {
@@ -66,54 +65,54 @@ public class SecurityAuditRestController {
 			return SecurityLogResponseDTO.builder().totalLogs(1).currentLogs(1).securityAuditLogs(securityAuditLogsDTO)
 			        .totalPages(1).currentPage(0).build();
 		}
-		
+
 		if (page < 0) {
 			page = 0;
 		}
 		if (size <= 0) {
 			size = 15;
 		}
-		
+
 		Date start = UtilClass.parseDate(startDate, false);
 		Date end = UtilClass.parseDate(endDate, true);
-		
+
 		List<AuditSecurityEvent> securityEvents = auditService.getSecurityEvents(eventType, username, start, end, page,
 		    size);
 		long totalCount = auditService.countSecurityEvents(eventType, username, start, end);
 		int totalPages = UtilClass.computeTotalPages(totalCount, size);
-		
+
 		List<SecurityAuditLogDTO> securityAuditLogsDTO = mapToDTOs(securityEvents);
-		
+
 		return SecurityLogResponseDTO.builder().totalLogs(totalCount).currentLogs(securityAuditLogsDTO.size())
 		        .totalPages(totalPages).currentPage(page).securityAuditLogs(securityAuditLogsDTO).build();
 	}
-	
+
 	@GetMapping("/relatedAudits")
 	public SecurityLogResponseDTO fetchRelatedAudits(@RequestParam(value = "sessionId") String sessionId,
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
-		
+
 		if (sessionId == null || sessionId.isEmpty()) {
 			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
 		}
-		
+
 		if (page < 0) {
 			page = 0;
 		}
 		if (size <= 0) {
 			size = 15;
 		}
-		
+
 		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, page, size);
 		long totalCount = auditService.countRelatedSecurityEvents(sessionId);
 		int totalPages = UtilClass.computeTotalPages(totalCount, size);
-		
+
 		List<SecurityAuditLogDTO> securityAuditLogsDTO = mapToDTOs(allRelated);
-		
+
 		return SecurityLogResponseDTO.builder().totalLogs(totalCount).currentLogs(securityAuditLogsDTO.size())
 		        .totalPages(totalPages).currentPage(page).securityAuditLogs(securityAuditLogsDTO).build();
 	}
-	
+
 	private SecurityAuditLogDTO mapToDTO(AuditSecurityEvent event) {
 		if (event == null) {
 			return null;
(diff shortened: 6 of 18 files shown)


sudhanshu-raj:
hi @ManojLL @wikumChamith , let me know your thoughts

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+		int totalPages = UtilClass.computeTotalPages(totalCount, size);
+		
+		int fromIndex = page * size;
+		List<AuditSecurityEvent> pagedList;
+		if (fromIndex >= allRelated.size()) {
+			pagedList = Collections.emptyList();
+		} else {
+			int toIndex = Math.min(fromIndex + size, allRelated.size());
+			pagedList = allRelated.subList(fromIndex, toIndex);
+		}
```
dkayiwa: Paging this in memory needs to change before merge, for two reasons.

`page * size` is int arithmetic on an unclamped request parameter, so a large page overflows. Running it against this controller with `RestExceptionHandler` in place:

```
GET /rest/v1/securityauditlogs/releatedAudits?sessionId=SESSION-A&page=200000000&size=15
-> 500 {"error":"Internal Server Error","message":"An unexpected error occurred"}
```

`page * size` is -1294967296, which slips past the `fromIndex >= allRelated.size()` guard, and then `subList` throws `IndexOutOfBoundsException`. The main endpoint survives the same input because Hibernate rejects a negative `setFirstResult` as an `IllegalArgumentException` (400), so this is the only path that 500s.

The other problem is that `totalLogs` and `totalPages` are derived from a list the DAO has already truncated. `getRelatedSecurityEvents` ends in `setMaxResults(limit)`, so with the hardcoded 1000 a session with more events than that reports `totalLogs: 1000` and the older rows are unreachable at any page. Nothing throws, so a client paging through just gets a quietly wrong total.

`ReadAuditRestController.fetchRelatedAudits` already has the shape I would copy: `getRelatedReadLogs(sessionId, page, size)` alongside `countRelatedReadLogs(sessionId)`. Adding the matching pair to `AuditService`/`AuditDao` for security events fixes both problems and gets rid of the `1000` literal at the same time. `UtilClass.paginate` is not a shortcut here, by the way, it has the same negative-index hole.

omod/src/test/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestControllerTest.java
```diff
@@ -0,0 +1,131 @@
...
+	@Test
+	public void shouldFetchSecurityAuditsSuccessfullyWithIdFilter() throws Exception {
+		AuditSecurityEvent mockEvent = mock(AuditSecurityEvent.class);
+		when(mockEvent.getEventType()).thenReturn(AuditSecurityEventType.LOGIN_SUCCESS);
+		when(auditService.getSecurityEventById(1)).thenReturn(mockEvent);
+		
+		mockMvc.perform(get("/rest/v1/securityauditlogs").param("logId", "1")).andExpect(status().isOk());
+		
+		verify(auditService).getSecurityEventById(1);
+	}
```
dkayiwa, line 91: I'd like a success-path assertion here before this merges. The tests only check the status code, and with `mock(AuditSecurityEvent.class)` every getter except `getEventType()` returns null, so the field mapping in `mapToDTO` is never exercised. I swapped `username` with `userUuid` and `ipAddress` with `userAgent` in `mapToDTO` and all seven tests still passed, which means a misattributed IP or user in the audit payload would ship with green CI.

`AuditSecurityEvent` has a Lombok builder, so no mock is needed:

```java
AuditSecurityEvent event = AuditSecurityEvent.builder().id(7).eventType(AuditSecurityEventType.LOGIN_SUCCESS)
        .username("admin").userUuid("user-uuid-1").eventTime(new Date()).ipAddress("10.0.0.7")
        .userAgent("curl/8.4").sessionId("SESSION-A").details("{}").build();
```

Asserting the JSON fields off that, plus `totalLogs`/`currentLogs`/`totalPages`/`currentPage`, would close the gap.

Worth stretching `shouldFetchRelatedAuditsSuccessfully` too: the paging arithmetic in `fetchRelatedAudits` is the only genuinely new logic in this PR and the test covers a single-element first page only. A case with more items than fit on a page, and one requesting a page past the end, would have caught the overflow I mentioned on the controller.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+
+@RestController
+@RequestMapping("/rest/" + RestConstants.VERSION_1 + "/securityauditlogs")
+@RequiredArgsConstructor
+public class SecurityAuditRestController {
+	
+	private final AuditService auditService;
```
dkayiwa, line 37: This one is worth fixing but I don't think it blocks the PR: a caller who lacks `View Security Audit Logs` gets a 500 from these endpoints instead of a 403. The service methods are `@Authorized(VIEW_SECURITY_AUDIT_LOGS)` and the injected proxy runs core's `AuthorizationAdvice`, so an unprivileged call throws `APIAuthenticationException`, and `RestExceptionHandler` has no branch for it. It falls through to `handleGeneralError`:

```
500 {"error":"Internal Server Error","message":"An unexpected error occurred"}
```

I checked that with this controller and that advice in a standalone MockMvc setup. For comparison, webservices.rest's own `BaseRestController.apiAuthenticationExceptionHandler` answers 403, and nothing else is competing to handle it here since webservices.rest 2.49.0 ships no `@ControllerAdvice` of its own.

An `@ExceptionHandler(APIAuthenticationException.class)` returning 403 in `RestExceptionHandler` would cover this controller and the three that already exist. Permission denials on a security-audit endpoint are the ones most likely to be hit in practice, and 500s there will send people hunting for a server fault that isn't there.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+		
+		if (logId != null) {
+			AuditSecurityEvent securityEvent = auditService.getSecurityEventById(logId);
+			if (securityEvent == null) {
+				return SecurityLogResponseDTO.builder().totalLogs(0).currentLogs(0)
+				        .securityAuditLogs(Collections.emptyList()).build();
+			}
+			List<SecurityAuditLogDTO> securityAuditLogsDTO = mapToDTOs(Collections.singletonList(securityEvent));
+			return SecurityLogResponseDTO.builder().totalLogs(1).currentLogs(1).securityAuditLogs(securityAuditLogsDTO)
+			        .build();
```
dkayiwa: A single-log lookup reports `totalPages: 0` even though it returns a row:

```
GET /rest/v1/securityauditlogs?logId=7
{"totalLogs":1,"currentLogs":1,"totalPages":0,"currentPage":0,"securityAuditLogs":[{...}]}
```

A client driving its loop off `totalPages` would skip the record it just asked for.

```suggestion
			return SecurityLogResponseDTO.builder().totalLogs(1).currentLogs(1).totalPages(1).currentPage(0)
			        .securityAuditLogs(securityAuditLogsDTO).build();
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+@RestController
+@RequestMapping("/rest/" + RestConstants.VERSION_1 + "/securityauditlogs")
+@RequiredArgsConstructor
+public class SecurityAuditRestController {
+	
+	private final AuditService auditService;
+	
+	@GetMapping
+	public SecurityLogResponseDTO fetchSecurityAudits(@RequestParam(value = "logId", required = false) Integer logId,
+	        @RequestParam(value = "eventType", required = false) String eventType,
```
dkayiwa, line 41: Not a blocker, but a misspelled `eventType` comes back as an empty result rather than an error. `AuditSecurityEventType.fromName` maps anything unrecognised to `UNKNOWN`, and the DAO then filters on that enum value, so `?eventType=LOGIN_SUCESS` returns `{"totalLogs":0,...}` and the caller can't tell a typo from a genuinely empty range. `UNKNOWN` is a real stored value too, so `?eventType=UNKNOWN` and `?eventType=garbage` are indistinguishable.

This method already rejects a malformed `startDate` with a 400, so validating `eventType` against the enum and throwing `IllegalArgumentException` listing the accepted values would be consistent and much friendlier to API clients.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+		long totalCount = auditService.countSecurityEvents(eventType, username, start, end);
+		int totalPages = UtilClass.computeTotalPages(totalCount, size);
+		
+		List<SecurityAuditLogDTO> securityAuditLogsDTO = mapToDTOs(securityEvents);
+		
+		return SecurityLogResponseDTO.builder().totalLogs(totalCount).currentLogs(securityAuditLogsDTO.size())
+		        .totalPages(totalPages).currentPage(page).securityAuditLogs(securityAuditLogsDTO).build();
+	}
+	
+	@GetMapping("/releatedAudits")
```
dkayiwa: Not a correctness problem, but `releatedAudits` is misspelled, and the window to fix it for free closes at the next release. `ReadAuditRestController` carries the same typo and neither endpoint has shipped: both landed after the `auditlogweb-1.0.0` tag and are still on 1.1.0-SNAPSHOT. Renaming both to `relatedAudits` now costs nothing, whereas after a release it's a permanent part of the URL contract.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: I thought of  avoiding inner changes but that seems to be bad idea. anyway fixed it with the same thought.

omod/src/test/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestControllerTest.java
sudhanshu-raj: Modified with the more robust test cases.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: fixed

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: fixed, checking this early now

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: done, added it

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: fixed!

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
```diff
@@ -334,10 +334,14 @@ void logSecurityEvent(AuditSecurityEventType eventType, String username, String
...
-	 * @param limit the maximum number of events to return
+	 * @param page defines the particular page in a paginated view
+	 * @param size defines how many logs want in one go
 	 * @return a list of {@link AuditSecurityEvent} ordered by eventTime descending
 	 */
 	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
-	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int limit);
+	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int size, int page);
+	
+	@Authorized(AuditLogConstants.VIEW_READ_AUDIT_LOGS)
```
dkayiwa: The privilege on this method is wrong, and it takes the `/relatedAudits` endpoint down with it: it is gated on `VIEW_READ_AUDIT_LOGS`, while every other security-audit method in this interface, including `getRelatedSecurityEvents` two lines up, uses `VIEW_SECURITY_AUDIT_LOGS`.

If merged as-is, a user granted only "View Security Audit Logs" cannot use `GET /rest/v1/securityauditlogs/relatedAudits` at all, because `fetchRelatedAudits` calls both service methods in sequence: the first passes the authorization advice, the second throws `APIAuthenticationException`. I wrapped the service in core's `AuthorizationAdvice` the way `moduleApplicationContext.xml`'s `serviceInterceptors` does, with a user context holding the security privilege but not the read one:

```
GET /rest/v1/securityauditlogs/relatedAudits?sessionId=SESSION-A
  -> 403 {"error":"Unauthorized access", ...}
GET /rest/v1/securityauditlogs
  -> 200
```

config.xml declares the three privileges separately and no role is handed them together, so a security-auditor role is exactly the case that breaks. It goes the other way too: someone with only "View Read Audit Logs" can call this and learn how many security events a session has.

```suggestion
	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
```

Nothing catches it today because the controller tests mock `AuditService`, so the advice never runs. `ServiceAuthorizationTest` is where the two sibling checks live and would be a natural home for this one.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
```diff
@@ -334,10 +334,14 @@ void logSecurityEvent(AuditSecurityEventType eventType, String username, String
...
 	 *
 	 * @param sessionId the session ID to filter by
-	 * @param limit the maximum number of events to return
+	 * @param page defines the particular page in a paginated view
+	 * @param size defines how many logs want in one go
 	 * @return a list of {@link AuditSecurityEvent} ordered by eventTime descending
 	 */
 	@Authorized(AuditLogConstants.VIEW_SECURITY_AUDIT_LOGS)
-	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int limit);
+	List<AuditSecurityEvent> getRelatedSecurityEvents(String sessionId, int size, int page);
```
dkayiwa: The parameter order here is inverted relative to the Javadoc directly above it, which lists `@param page` before `@param size`, and relative to `ReadAuditService.getRelatedReadLogs(String sessionId, int page, int size)`. Because `page * size` is commutative the offset comes out right either way, so a caller who writes the arguments in the documented order gets no error at all, just a wrongly sized page. Capturing what actually reaches the query:

```
dao.getRelatedSecurityEvents("S", 0, 15)    // meaning page 0, size 15
  -> setFirstResult(0)   setMaxResults(0)     // asks the DB for zero rows
dao.getRelatedSecurityEvents("S", 2, 15)    // meaning page 2, size 15
  -> setFirstResult(30)  setMaxResults(2)     // right offset, 2 rows instead of 15
```

Both current callers pass it correctly so nothing is broken right now, but this is a public service API and the window to change it for free closes at the next release. I would swap it to `(String sessionId, int page, int size)` across `AuditService`, `AuditServiceImpl` and `AuditDao` so it matches its own Javadoc and the read-audit sibling. Two small things while you are in those files: `countRelatedSecurityEvents` below is the only method in this interface with no Javadoc, and the DAO spells it `countRelatedSecurityEvent`, singular, where the sibling is `countRelatedReadLogs`.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,136 @@
...
+	        @RequestParam(value = "startDate", required = false) String startDate,
+	        @RequestParam(value = "endDate", required = false) String endDate,
+	        @RequestParam(value = "page", defaultValue = "0") int page,
+	        @RequestParam(value = "size", defaultValue = "15") int size) {
+		
+		if (logId != null && logId <= 0) {
+			throw new IllegalArgumentException("Please provide a valid log ID");
+		}
+		
+		if (eventType != null) {
```
dkayiwa: This rejects a blank `eventType` along with a bad one, which I do not think is the intent. Both an empty and a whitespace value come back 400, while the equivalent blank `username` is treated as no filter:

```
GET /rest/v1/securityauditlogs?eventType=    -> 400 {"error":"Bad Request","message":"Invalid eventType "}
GET /rest/v1/securityauditlogs?username=     -> 200 (unfiltered)
```

`fromName` returns null for blank input precisely so that blank means "no filter", and `bindSecurityEventFilters` skips a blank `username` for the same reason. A UI that always emits its filter parameters will now get a 400 whenever the event-type dropdown is left empty. Skipping the check for a blank value keeps this consistent with how `username` behaves:

```suggestion
		if (eventType != null && !eventType.trim().isEmpty()) {
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,6 +85,11 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
 		return buildResponseEntity("Audit Log Unavailable", ex.getMessage(), HttpStatus.SERVICE_UNAVAILABLE);
 	}
 	
+	@ExceptionHandler(APIAuthenticationException.class)
+	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
+		return buildResponseEntity("Unauthorized access", ex.getMessage(), HttpStatus.FORBIDDEN);
+	}
```
dkayiwa, line 102: An unauthenticated caller reaches this handler too, and for that case the platform answers 401 rather than 403. In webservices.rest 2.49.0, `BaseRestController.apiAuthenticationExceptionHandler` branches on `Context.isAuthenticated()`: logged in but missing the privilege gives 403, not logged in gives 401 plus a `WWW-Authenticate: Basic realm=...` header. Its `AuthorizationFilter` deliberately lets credential-less requests straight through ("It will not fail on invalid or missing credentials. We count on the API to throw exceptions if an unauthenticated user tries to do something they are not allowed to do"), so the unauthenticated case genuinely lands here instead of being bounced upstream. As written, every caller gets 403, so a client whose session expired cannot tell "log in again" from "you will never be allowed to see this". My suggestion last round to return 403 was only half the story.

```suggestion
	@ExceptionHandler(APIAuthenticationException.class)
	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
		if (Context.isAuthenticated()) {
			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
		}
		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
	}
```

That needs an `org.openmrs.api.context.Context` import. I also changed the `error` labels: the other handlers here use the status reason phrase, and "Unauthorized access" sitting on a 403 reads like a 401. Sending the `WWW-Authenticate` challenge as well would match `BaseRestController` exactly, but the status code is the part that matters.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
sudhanshu-raj: Fixed, I ordered the doc lines

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
sudhanshu-raj: Ahh I missed it, fixed now

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
sudhanshu-raj: thanks, fixed !

omod/src/test/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestControllerTest.java
```diff
@@ -0,0 +1,180 @@
...
+		        .andExpect(jsonPath("$.error", is("Bad Request")))
+		        .andExpect(jsonPath("$.message", is("Please provide a valid log ID")));
+	}
+	
+	@Test
+	public void shouldThrowErrorIfInvalidEventTypePassed() throws Exception {
+		mockMvc.perform(get("/rest/v1/securityauditlogs").param("eventType", "LOGIN_SUCESS"))
+		        .andExpect(status().isBadRequest()).andExpect(jsonPath("$.error", is("Bad Request")))
+		        .andExpect(jsonPath("$.message", is("Invalid eventType LOGIN_SUCESS")));
+	}
```
dkayiwa, line 113: The two behavior changes in the last couple of commits both sit in code this class already reaches, and neither has a test.

A blank `eventType` now means "no filter" instead of a 400, which is the empty-dropdown case, and nothing here covers it:

```java
mockMvc.perform(get("/rest/v1/securityauditlogs").param("eventType", "")).andExpect(status().isOk());
```

The 401 vs 403 split in `RestExceptionHandler` has no test anywhere in the repo. All four REST controller tests wire the advice in, but none of them ever throws `APIAuthenticationException`, so neither branch is exercised. It is reachable from this exact setup, since `mockito-inline` is already on the omod test classpath:

```java
when(auditService.getSecurityEvents(isNull(), isNull(), isNull(), isNull(), anyInt(), anyInt()))
        .thenThrow(new APIAuthenticationException("Privileges required: View Security Audit Logs"));
try (MockedStatic<Context> ctx = mockStatic(Context.class)) {
    ctx.when(Context::isAuthenticated).thenReturn(false);
    mockMvc.perform(get("/rest/v1/securityauditlogs")).andExpect(status().isUnauthorized())
            .andExpect(jsonPath("$.error", is("Unauthorized")));
}
```

I ran both of those against this branch and they pass, so it is mostly a matter of committing them. The behavior is right today so this is not a correctness blocker, but I would land it in the same PR: the wrong privilege constant got as far as it did here precisely because the service is mocked out of every controller test, and `ServiceAuthorizationTest` still has no case for `countRelatedSecurityEvents`.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
dkayiwa: Both callers pass this correctly so nothing is broken today, but the two sibling methods still take their int pair in opposite orders: `getRelatedSecurityEvents(String sessionId, int size, int page)` here, and `ReadAuditService.getRelatedReadLogs(String sessionId, int page, int size)` next door. Both are `(String, int, int)`, so a call written in the wrong order compiles and quietly pages wrong. Run against a real database with 80 events in one session:

```
getRelatedSecurityEvents("S", 15, 0)  -> 15 rows   (size 15, page 0)
getRelatedSecurityEvents("S", 15, 2)  -> 15 rows   (size 15, page 2)
getRelatedSecurityEvents("S",  0, 15) ->  0 rows   (what someone writing "page 0, size 15" gets)
getRelatedSecurityEvents("S",  2, 15) ->  2 rows   (what someone writing "page 2, size 15" gets)
```

I would still swap it to `(String sessionId, int page, int size)`. The signature is already changing in this PR, it was `(sessionId, limit)` on main, so matching the read-audit sibling costs the same three call sites now and becomes a breaking change once 1.1.0 ships.

The other half of this is that nothing in the build pins the offset. I changed `AuditDao.getRelatedSecurityEvents` to `query.setFirstResult(page)` and ran the lot: 196 api tests and 67 omod tests, all green. `AuditDaoTest` only ever asks for page 0, where `page * size` and `page` agree. A case like `getRelatedSecurityEvents("s", 15, 2)` asserting `verify(query).setFirstResult(30)` would close it.

sudhanshu-raj: Done, and tweaked the existing test case for this  case.

omod/src/test/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestControllerTest.java
sudhanshu-raj: Thanks for suggesting the code, yeah putting these more test cases will help in test more edge cases. And on `ServiceAuthorizationTest ` is meant to verify the AOP wiring where it would verify the  one test per privilege  for `@Authorized` interceptor and verifying one such interceptor for a privilege is enough for it, that's  why I don't think  covering every method under the same privilege is necessary.

sudhanshu-raj:
> hi @ManojLL @wikumChamith , let me know your thoughts

@wikumChamith @ManojLL ,

dkayiwa:
@claude review

omod/src/main/java/org/openmrs/module/auditlogweb/web/controller/SecurityAuditDetailController.java
```diff
@@ -83,7 +83,7 @@ public ModelAndView showDetails(HttpServletRequest request, ModelMap model) {
 			// Fetch related events from the same session
 			List<AuditSecurityEvent> relatedEvents = null;
 			if (event.getSessionId() != null && !event.getSessionId().isEmpty()) {
-				relatedEvents = auditService.getRelatedSecurityEvents(event.getSessionId(), RELATED_EVENTS_LIMIT);
+				relatedEvents = auditService.getRelatedSecurityEvents(event.getSessionId(), RELATED_EVENTS_LIMIT, 0);
```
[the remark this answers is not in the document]
sudhanshu-raj: fixed !

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,136 @@
...
+		}
+		
+		if (page < 0) {
+			page = 0;
+		}
+		if (size <= 0) {
+			size = 15;
+		}
+		
+		List<AuditSecurityEvent> allRelated = auditService.getRelatedSecurityEvents(sessionId, size, page);
```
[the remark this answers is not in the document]
sudhanshu-raj: fixed !

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditService.java
sudhanshu-raj: fixed !

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,136 @@
...
+			AuditSecurityEventType parsed = AuditSecurityEventType.fromName(eventType);
+			if (parsed == null || parsed == AuditSecurityEventType.UNKNOWN) {
+				throw new IllegalArgumentException("Invalid eventType " + eventType);
+			}
+		}
+		
+		if (logId != null) {
+			AuditSecurityEvent securityEvent = auditService.getSecurityEventById(logId);
+			if (securityEvent == null) {
+				return SecurityLogResponseDTO.builder().totalLogs(0).currentLogs(0).totalPages(0).currentPage(0)
```
wikumChamith: I think it would be better if we throw a `ResponseStatusException` error here. Currently this is just gonna return a 200 with an empty page.

sudhanshu-raj: Done !

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityLogResponseDTO.java
```diff
@@ -0,0 +1,31 @@
...
+@Builder
+public class SecurityLogResponseDTO {
+	
+	private long totalLogs;
+	
+	private int currentLogs;
+	
+	private int totalPages;
+	
+	private int currentPage;
```
ManojLL, line 27: `currentPage` is already available in the pagination context. Is it necessary to send it back to the client?

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,136 @@
...
+			return SecurityLogResponseDTO.builder().totalLogs(1).currentLogs(1).securityAuditLogs(securityAuditLogsDTO)
+			        .totalPages(1).currentPage(0).build();
+		}
+		
+		if (page < 0) {
+			page = 0;
+		}
+		if (size <= 0) {
+			size = 15;
+		}
```
ManojLL: Would it be better to add validation directly to the request parameters, such as `@Min(0)` for page and `@Positive` for size, instead of handling these validations manually?

ManojLL (changes requested):

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: Actually using these annotations completely flips the situation means here if page is negative then we are setting it to 0 by default and if size <= 0 then take as 15 by default but these annotations will just validate these values and if not verifies then throw the exception or simple the error body instead of the default results we returning currently. 

Second these annotations  will probably not the best case if we need to throw the bad request body because first we need to register the the `MethodValidationPostProcessor`  bean and add `@Validator` on controller class and then most hurdle will come on testing this because we using core 2.7.x which using spring 5.3.x so to test that we need to manually create the proxy  for test case because `MockMvc`  standaloneSetup not supports proxy for this spring version and manually creating is more mess instead if our goal is to validate or throw exception we can manually throw using the `IllegalArgumentException`.

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityLogResponseDTO.java
sudhanshu-raj: Yeah that would be redundant if the client already sets the page parameter in the request then we are going to return same on the response too, but if client not sets page parameter then in that case it would be sightly useful in the response the client will have clear picture of from which page this logs belongs too. So it guess it's not that unnecessary, otherwise we can remove it.

sudhanshu-raj:
Hey @wikumChamith @ManojLL @dkayiwa,  please let me know if any further things I need to modify.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
```diff
@@ -81,7 +81,7 @@ public ReadAuditLogResponseDTO fetchReadAudits(@RequestParam(value = "logId", re
 		
 	}
 	
-	@GetMapping("/releatedAudits")
+	@GetMapping("/relatedAudits")
```
dkayiwa, line 81: While you're in this file renaming the endpoint: `fetchReadAudits` above still has the two `logId` behaviours that got changed in the security controller this round, so the twins now disagree. I ran both against this branch:

```
GET /rest/v1/readauditlogs?logId=7      -> 200 {"totalLogs":1,"currentLogs":1,"totalPages":0,...}
GET /rest/v1/securityauditlogs?logId=7  -> 200 {"totalLogs":1,"currentLogs":1,"totalPages":1,...}

GET /rest/v1/readauditlogs?logId=999     -> 200 {"totalLogs":0,"currentLogs":0,"totalPages":0,"readAuditLogs":[]}
GET /rest/v1/securityauditlogs?logId=999 -> 404 {"error":"Not Found","message":"No log found for this logId"}
```

A client driving its loop off `totalPages` skips the read log it just asked for, and one that checks for a 404 to spot a missing id never gets one from `/readauditlogs`. `ReadAuditRestControllerTest` asserts neither of these (`shouldFetchReadAuditsSuccessfullyWithIdFilter` only checks the status code, and there is no missing-id case at all), which is why it stayed invisible.

Neither endpoint has been released, the only tag is `auditlogweb-1.0.0` from February and both landed on 1.1.0-SNAPSHOT, so straightening the read side out is still free. It is the same two edits you already made in `SecurityAuditRestController`: add `.totalPages(1).currentPage(0)` to the found branch on line 59, and throw `ResponseStatusException(HttpStatus.NOT_FOUND, ...)` on line 54 instead of returning an empty page. I would rather it rode along here since you are already in the file, but a separate ticket is fine too. Not a blocker either way.

sudhanshu-raj: Since file got touched, let me fix this. And done !

dkayiwa: The `totalPages` half of this is still open, so the two endpoints still disagree on a found id:

```
GET /rest/v1/readauditlogs?logId=7      -> {"totalLogs":1,"currentLogs":1,"totalPages":0,"currentPage":0,"readAuditLogs":[{"id":7,...}]}
GET /rest/v1/securityauditlogs?logId=7  -> {"totalLogs":1,"currentLogs":1,"totalPages":1,"currentPage":0,"securityAuditLogs":[{"id":7,...}]}
```

I ran both against 60c9888 using the same standalone MockMvc setup as the controller tests. A client that loops while `page < totalPages` runs zero iterations, so it drops the record it just fetched by id.

Line 58 needs the pair you already added on the security side:

```java
return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).totalPages(1).currentPage(0)
        .readAuditLogs(readAuditLogsDTO).build();
```

Adding `totalPages` to the assertions in `shouldFetchReadAuditsSuccessfullyWithIdFilter` would pin it, that test still only checks the status code. Not a blocker, same as last round.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,6 +86,14 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
 		return buildResponseEntity("Audit Log Unavailable", ex.getMessage(), HttpStatus.SERVICE_UNAVAILABLE);
 	}
 	
+	@ExceptionHandler(APIAuthenticationException.class)
+	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
+		if (Context.isAuthenticated()) {
+			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
+		}
+		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
+	}
```
wikumChamith, line 102: Let's log here so we can see why the request fails here.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,135 @@
...
```
wikumChamith: `size` has no upper bound, so `?size=500000` would load and serialize every matching row in a single request. I think we should cap it to a max value.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,135 @@
...
+		return SecurityLogResponseDTO.builder().totalLogs(totalCount).currentLogs(securityAuditLogsDTO.size())
+		        .totalPages(totalPages).currentPage(page).securityAuditLogs(securityAuditLogsDTO).build();
+	}
+	
+	@GetMapping("/relatedAudits")
+	public SecurityLogResponseDTO fetchRelatedAudits(@RequestParam(value = "sessionId") String sessionId,
+	        @RequestParam(value = "page", defaultValue = "0") int page,
+	        @RequestParam(value = "size", defaultValue = "15") int size) {
+		
+		if (sessionId == null || sessionId.isEmpty()) {
```
wikumChamith:
```suggestion
		if (sessionId == null || sessionId.trim().isEmpty()) {
```

wikumChamith (commented):
Few minor nits

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: Added the cap limit to 200

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,6 +90,16 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
...
 	
+	@ExceptionHandler(APIAuthenticationException.class)
+	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
+		if (Context.isAuthenticated()) {
+			log.warn("Request failed due failed authentication attempt");
+			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
+		}
+		log.warn("Request failed due to not have enough permissions");
+		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
+	}
```
wikumChamith, line 102: Shouldn't these log messages be swapped?

api/src/main/java/org/openmrs/module/auditlogweb/api/utils/AuditLogConstants.java
```diff
@@ -17,4 +17,6 @@ public final class AuditLogConstants {
 	public static final String VIEW_SECURITY_AUDIT_LOGS = "View Security Audit Logs";
 	
 	public static final String VIEW_READ_AUDIT_LOGS = "View Read Audit Logs";
+	
+	public static final Integer MAX_PAGE_SIZE = 200;
```
wikumChamith: Let's drop the wrapper and use a primitive int here.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
```diff
@@ -51,12 +52,12 @@ public ReadAuditLogResponseDTO fetchReadAudits(@RequestParam(value = "logId", re
...
-				return ReadAuditLogResponseDTO.builder().totalLogs(0).currentLogs(0).readAuditLogs(Collections.emptyList())
-				        .build();
+				throw new ResponseStatusException(HttpStatus.NOT_FOUND, "No log found for this logId");
 			}
 			List<ReadAuditLogDTO> readAuditLogsDTO = readAuditService
 			        .mapToReadAuditLogDTO(Collections.singletonList(readAuditLog));
-			return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).readAuditLogs(readAuditLogsDTO).build();
+			return ReadAuditLogResponseDTO.builder().totalLogs(1).currentLogs(1).readAuditLogs(readAuditLogsDTO)
+			        .totalPages(1).currentPage(0).build();
 		}
```
wikumChamith, line 61: Why the duplicated `currentPage(0)`?

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,6 +90,16 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
...
+	
 	@ExceptionHandler(Exception.class)
 	public ResponseEntity<Map<String, String>> handleGeneralError(Exception ex) {
```
wikumChamith, line 105: Not related to this, but I think we should also add a log line here in this PR.

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
```diff
@@ -0,0 +1,40 @@
...
+	
+	private Integer id;
+	
+	private AuditSecurityEventType eventType;
+	
+	private String username;
+	
+	private String userUuid;
+	
+	private Date eventTime;
```
wikumChamith, line 32:
```suggestion
	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd/MM/yyyy HH:mm:ss", timezone = "GMT")
	private Date eventTime;
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,137 @@
...
+			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
+		}
+		
+		if (page < 0) {
+			page = 0;
+		}
+		if (size <= 0) {
+			size = 15;
+		}
+
```
wikumChamith: Don't we need to clamp to the max page size here?

wikumChamith (commented):
Apart from these points, I'm also seeing some outdated comments. @sudhanshu-raj can we update them?

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
sudhanshu-raj: Added for the read controller too

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
sudhanshu-raj: Seems not needed, done!

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
sudhanshu-raj: My bad, swapped

sudhanshu-raj:
@claude review

omod/src/test/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestControllerTest.java
```diff
@@ -71,7 +74,9 @@ public void shouldFetchReadAuditsSuccessfullyWithIdFilter() throws Exception {
 		when(readAuditService.getReadAuditLogById(1)).thenReturn(mockReadAuditLog);
 		when(readAuditService.mapToReadAuditLogDTO(logList)).thenReturn(Collections.emptyList());
 		
-		mockMvc.perform(get("/rest/v1/readauditlogs").param("logId", "1")).andExpect(status().isOk());
+		mockMvc.perform(get("/rest/v1/readauditlogs").param("logId", "1")).andExpect(status().isOk())
+		        .andExpect(jsonPath("$.totalLogs", is(1))).andExpect(jsonPath("$.currentLogs", is(1)))
+		        .andExpect(jsonPath("$.totalPages", is(1))).andExpect(jsonPath("$.currentPage", is(0)));
```
dkayiwa: The format change to `eventTime` never reaches a test. `ReadAuditLogDTO` picks up `@JsonFormat(pattern = "dd/MM/yyyy HH:mm:ss", timezone = "GMT")` in this PR, so `/readauditlogs` goes from returning `eventTime` as epoch milliseconds on main to returning a formatted string, but every test in this class stubs `mapToReadAuditLogDTO` to `Collections.emptyList()`, so no `ReadAuditLogDTO` is ever serialized. I swapped the pattern for a nonsense string and ran the whole build: 196 api and 76 omod tests, all green.

`SecurityAuditRestControllerTest` already pins its own side with `jsonPath("$.securityAuditLogs[0].eventTime", is("25/12/2026 14:30:00"))`, so this is really just mirroring that one assertion. I ran the version below (passes), and re-ran it against the broken pattern (fails with a 500 on the serialization), so it does close the gap:

```suggestion
		Date fixedDate = Date.from(LocalDateTime.of(2026, 12, 25, 14, 30, 0).atZone(ZoneId.of("GMT")).toInstant());
		when(readAuditService.mapToReadAuditLogDTO(logList)).thenReturn(
		    Collections.singletonList(ReadAuditLogDTO.builder().id(1).eventTime(fixedDate).build()));
		
		mockMvc.perform(get("/rest/v1/readauditlogs").param("logId", "1")).andExpect(status().isOk())
		        .andExpect(jsonPath("$.readAuditLogs[0].eventTime", is("25/12/2026 14:30:00")))
		        .andExpect(jsonPath("$.totalLogs", is(1))).andExpect(jsonPath("$.currentLogs", is(1)))
		        .andExpect(jsonPath("$.totalPages", is(1))).andExpect(jsonPath("$.currentPage", is(0)));
```

That needs `org.openmrs.module.auditlogweb.api.dto.ReadAuditLogDTO`, `java.time.LocalDateTime`, `java.time.ZoneId` and `java.util.Date` added to the imports. Not a blocker.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
```diff
@@ -81,7 +83,7 @@ public ReadAuditLogResponseDTO fetchReadAudits(@RequestParam(value = "logId", re
 		
 	}
 	
-	@GetMapping("/releatedAudits")
+	@GetMapping("/relatedAudits")
 	public ReadAuditLogResponseDTO fetchRelatedAudits(@RequestParam(value = "sessionId") String sessionId,
```
dkayiwa, line 82: While the two `/relatedAudits` endpoints are being lined up, this one still guards with `sessionId.isEmpty()` on line 91 where the security controller now uses `sessionId.trim().isEmpty()`, so a whitespace-only session id is a bad request on one and an empty page on the other. Both driven through the same standalone MockMvc setup the tests use:

```
GET /rest/v1/readauditlogs/relatedAudits?sessionId=%20      -> 200 {"totalLogs":0,"currentLogs":0,"totalPages":0,"currentPage":0,"readAuditLogs":[]}
GET /rest/v1/securityauditlogs/relatedAudits?sessionId=%20  -> 400 {"error":"Bad Request","message":"Invalid session id"}
```

`?sessionId=` with nothing after it is 400 on both, it is only the whitespace case that splits. Adding `trim()` on line 91 lines them up, and since neither endpoint has shipped the error contract is still free to settle. I ran the change and the suite stays green. Not a blocker.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,8 +90,20 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
...
 	
+	@ExceptionHandler(APIAuthenticationException.class)
+	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
+		if (Context.isAuthenticated()) {
+			log.warn("Request failed due to not have enough permissions");
+			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
+		}
+		
+		log.warn("Request failed due failed authentication attempt");
+		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
```
dkayiwa: Now that the two branches are the right way round, the wording is worth a second pass. "Request failed due to not have enough permissions" and "Request failed due failed authentication attempt" do not quite read as English, and neither carries the exception message, so the WARN that fires on every unauthenticated call to these endpoints tells a reader nothing they can act on.

```suggestion
		if (Context.isAuthenticated()) {
			log.warn("Request denied, the user does not have the required privileges: {}", ex.getMessage());
			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
		}
		
		log.warn("Request denied, the caller is not authenticated: {}", ex.getMessage());
		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
```

I deliberately kept it to the strings rather than reaching for `Context.getAuthenticatedUser()`, which would NPE under the `mockStatic(Context.class)` tests. Ran it, still green. Take or leave.

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -45,6 +49,8 @@
 @ControllerAdvice
```
wikumChamith: An unscoped `@ControllerAdvice` gets registered in the root context, so it applies to every controller in OpenMRS. The new `APIAuthenticationException` handler would turn other modules' login redirects into JSON 401s. Let's scope it to this module.

```suggestion
@ControllerAdvice(basePackages = "org.openmrs.module.auditlogweb.rest")
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/exceptions/RestExceptionHandler.java
```diff
@@ -84,8 +90,20 @@ public ResponseEntity<Map<String, String>> handleAuditLogUnavailable(AuditLogUna
...
 	
+	@ExceptionHandler(APIAuthenticationException.class)
+	public ResponseEntity<Map<String, String>> handleAPIAuthException(APIAuthenticationException ex) {
+		if (Context.isAuthenticated()) {
+			log.warn("Request denied, the user does not have the required privileges: {}", ex.getMessage());
+			return buildResponseEntity("Forbidden", ex.getMessage(), HttpStatus.FORBIDDEN);
+		}
+		
+		log.warn("Request denied, the caller is not authenticated: {}", ex.getMessage());
+		return buildResponseEntity("Unauthorized", ex.getMessage(), HttpStatus.UNAUTHORIZED);
```
wikumChamith: This echoes the exact privilege name to unauthenticated callers, and the message describes a 403 condition on a 401 response. Let's use a fixed message here. Keeping `ex.getMessage()` on the 403 branch is fine.

```suggestion
              return buildResponseEntity("Unauthorized", "Authentication required", HttpStatus.UNAUTHORIZED);
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,138 @@
...
+			        .totalPages(1).build();
+		}
+		
+		if (page < 0) {
+			page = 0;
+		}
+		if (size <= 0) {
+			size = 15;
+		}
+		size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
```
wikumChamith: `size` is capped but `page` isn't, so `page * size` in the DAO overflows int for huge pages, giving Hibernate a negative offset.

```suggestion
              size = Math.min(size, AuditLogConstants.MAX_PAGE_SIZE);
              // keep page * size within int range; the query offset is computed as page * size
              page = Math.min(page, Integer.MAX_VALUE / size);
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/SecurityAuditRestController.java
```diff
@@ -0,0 +1,128 @@
...
+	
+	@GetMapping("/relatedAudits")
+	public SecurityLogResponseDTO fetchRelatedAudits(@RequestParam(value = "sessionId") String sessionId,
+	        @RequestParam(value = "page", defaultValue = "0") int page,
+	        @RequestParam(value = "size", defaultValue = "15") int size) {
+		
+		if (sessionId == null || sessionId.trim().isEmpty()) {
+			throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
+		}
+
```
wikumChamith: The trim only happens inside the validation, so the untrimmed value still goes to the query. That means `?sessionId=%20abc%20` passes validation and silently returns `totalLogs: 0`, since the DAO does an exact match. Also, the null check is unreachable: the param is required, so Spring rejects a missing one before the method runs.

```suggestion
              if (sessionId.trim().isEmpty()) {
                      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid session id");
              }
              sessionId = sessionId.trim();
```

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
```diff
@@ -81,21 +78,17 @@ public ReadAuditLogResponseDTO fetchReadAudits(@RequestParam(value = "logId", re
...
 	        @RequestParam(value = "page", defaultValue = "0") int page,
 	        @RequestParam(value = "size", defaultValue = "15") int size) {
 		
-		if (sessionId == null || sessionId.isEmpty()) {
```
wikumChamith, line 89: Same as the security controller: validated rimmed, queried untrimmed

api/src/main/java/org/openmrs/module/auditlogweb/api/dto/SecurityAuditLogDTO.java
```diff
@@ -0,0 +1,42 @@
...
+@Builder
+public class SecurityAuditLogDTO {
+	
+	private Integer id;
+	
+	private AuditSecurityEventType eventType;
+	
+	private String username;
+	
+	private String userUuid;
```
wikumChamith: This endpoint emits `userUuid` while the existing read-audit API (`ReadAuditLogDTO`) emits `userUUID` for the same concept, so we have two sibling endpoints with different casings for one field. Worth settling before merge, since it's a public REST contract that's painful to change once clients exist.

sudhanshu-raj:
Hi @wikumChamith @dkayiwa , do have any other suggestions ?

omod/src/main/java/org/openmrs/module/auditlogweb/rest/ReadAuditRestController.java
```diff
@@ -81,21 +78,17 @@ public ReadAuditLogResponseDTO fetchReadAudits(@RequestParam(value = "logId", re
...
 		}
 		
-		if (page < 0) {
-			page = 0;
-		}
-		if (size <= 0) {
-			size = 15;
-		}
+		size = UtilClass.sanitizePageSizeValue(size);
+		page = UtilClass.sanitizePageValue(page, size);
```
wikumChamith: This sessionId is still queried untrimmed.

wikumChamith (commented):
One minor nit. Fix it and we are good to go :)

PR #59, opened 2026-07-31
AUDIT-62: Create missing Envers audit tables from the module at startup

wikumChamith:

  - Add createMissingAuditTables() to AuditBackfillDao: creates revision_entity from the Envers revision-entity persister and clones each missing audit table from its base table's JDBC metadata (nullable columns, REV/REVTYPE, composite (PK, REV) primary key)
  - Discover Envers' dynamic join/middle audit tables via a persister scan (resolveAllAuditTableMappings), in addition to the existing `@Audited` annotation scan
  - Read the audit table prefix/suffix from the SessionFactory's merged settings with Envers' real default _AUD, replacing the hardcoded _audit fallback (also fixes the backfill on stock core)
  - Gate creation behind EnversUtils.isEnversEnabled() in AuditBackfillService and run it from the activator before the one-time backfill, each independently guarded
  - Add pure-input unit tests for the DDL builders, type-sizing, and affix-stripping helpers, plus service gating tests

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-62

  No? -> execute above command


wikumChamith:
@druchniewicz

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDao.java
```diff
@@ -117,6 +122,199 @@ public List<TableMapping> orderByAuditTableDependencies(List<TableMapping> mappi
...
+	        Set<String> existingTables, String quote) {
+		String revColumnType = dialect.getTypeName(Types.INTEGER);
+		String revTypeColumnType = dialect.getTypeName(Types.TINYINT);
+		
+		int created = 0;
+		for (TableMapping mapping : resolveAllAuditTableMappings()) {
+			if (existingTables.contains(mapping.auditTable.toLowerCase(Locale.ROOT))) {
+				continue;
+			}
+			if (!existingTables.contains(mapping.baseTable.toLowerCase(Locale.ROOT))) {
```
dkayiwa, line 281: This skip drops two audit tables that Envers really does write to, so it needs fixing before merge. `Location_LocationAttribute_AUD` and `OrderSet_OrderSetMember_AUD` are middle tables Envers creates for audited collections, and there is no `Location_LocationAttribute` table to clone, so `stripAuditAffixes` hands back a base name that does not exist and creation stops here. Saving a Location that carries any LocationAttribute (or an OrderSet with members) on an Envers-enabled 2.7/2.8 server then throws `Table '<db>.location_locationattribute_aud' doesn't exist` and the whole transaction rolls back, so the edit is lost. Nothing before that point looks wrong: one warn line about a missing base table, and `createMissingAuditTables()` still reports its full count.

Running the DAO against a real 128-table OpenMRS MySQL schema, it created 110 tables and skipped those two, while Hibernate's own schema tooling over the same Envers metadata created 112. An audited save of a Location with one attribute then succeeded on the Hibernate-generated schema (a row landed in `Location_LocationAttribute_AUD`) and failed on the schema this code produces.

I would generate the DDL from Envers' own metadata rather than reverse-deriving a base table name, since for these two there is no base table to derive from. That is what `EnversAuditTableInitializer` does in core 2.9 via `SchemaMigrator`, and `HibernateSessionFactoryBean.getMetadata()` has been public since 2.4, so that route is available on the 2.7.0 floor. The clone path is fine for everything else it creates, so if you would rather keep this shape, the narrower fix is to fall back to the audit persister's own columns when the derived base table is absent, the way `createRevisionTableIfMissing` already builds `revision_entity`.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDao.java
```diff
@@ -117,6 +122,199 @@ public List<TableMapping> orderByAuditTableDependencies(List<TableMapping> mappi
...
+		}
+		return created;
+	}
+	
+	/**
+	 * All base/audit table pairs Envers writes to: the {@code @Audited} entities from
+	 * {@link #resolveAuditedTableMappings()} plus the dynamic audit entities Envers registers for
+	 * collection/join middle tables (e.g. role_privilege), which the annotation scan cannot see.
+	 */
+	public List<TableMapping> resolveAllAuditTableMappings() {
```
dkayiwa, line 316: Was keeping this fuller list for creation only deliberate? `backfillExistingDataIfEnabled()` still goes through `resolveAuditedTableMappings()`, which on a stock 2.7 metamodel returns 103 of the 111 audit tables Envers registers. Five of the six real join tables in that gap (`role_privilege_AUD`, `role_role_AUD`, `user_role_AUD`, `location_tag_map_AUD`, `order_type_class_map_AUD`) have a base primary key mirrored into the audit table, so `backfillTable` would handle them as they stand. If the intent is that role and privilege assignments get no baseline revision then that is fine, I just want to check it is not accidental now that the complete list exists next door.

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -39,6 +39,22 @@ public class AuditBackfillService {
...
+	 * Creates any missing Envers audit tables if Envers is enabled; a no-op when all tables already
+	 * exist. Not gated behind a global property: the tables are required for the module to function at
+	 * all, creation is idempotent, and only tables that do not exist are touched.
+	 */
+	public void createMissingAuditTablesIfEnabled() {
+		if (!EnversUtils.isEnversEnabled()) {
+			log.info("Envers is disabled (hibernate.integration.envers.enabled != true); skipping audit table creation.");
+			return;
+		}
+		int created = auditBackfillDao.createMissingAuditTables();
```
dkayiwa: `createMissingAuditTables()` catches every per-table failure inside the DAO and returns only the number that succeeded, so this cannot tell "everything already existed" from "every CREATE was rejected". Both arrive as 0 and log nothing at all. On a server whose connection user lacks CREATE, every audited write then fails with nothing louder than a warn line from boot.

Returning the skipped/failed count next to the created count, and logging at error when it is non-zero, would make an incomplete audit schema visible at startup instead of at the first audited write.

api/src/test/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDaoTest.java
```diff
@@ -114,6 +115,75 @@ void shouldQuoteIdentifierAndEscapeEmbeddedQuoteCharacter() {
...
+		String sql = dao.buildCreateRevisionTableSql("revision_entity", "id", "integer not null auto_increment", columns,
+		    "`");
+		
+		assertEquals("CREATE TABLE `revision_entity` (`id` integer not null auto_increment, "
+		        + "`timestamp` bigint NOT NULL, `changedBy` integer, `changedOn` datetime, PRIMARY KEY (`id`))",
+		    sql);
+	}
+	
+	@Test
+	void shouldBuildCreateAuditTableSqlWithCompositePrimaryKey() {
```
dkayiwa, line 165: These new tests all assert generated SQL strings built from hand-made inputs, so they pin down the string shape but never that a database accepts the DDL or that Envers can write into what gets created. Nothing in the suite calls `createMissingAuditTables()`, `resolveAllAuditTableMappings()`, `readColumnDefinitions()` or `revisionIdColumnDdl()`.

Worth closing that gap in this PR, because the PR itself makes it cheap: dropping `Context.getRuntimeProperties()` from the DAO leaves `AuditBackfillDao` constructible from a bare `SessionFactory`, so a test can boot Hibernate over two or three `@Audited` entities against H2, create only the base tables, call `createMissingAuditTables()`, then save an audited entity and read the revision back. Make one of those entities own a collection that Envers maps to a middle table.

wikumChamith:
@dkayiwa I've updated the PR.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDao.java
dkayiwa: The sixth one does need fixing before this merges: `concept_name_tag_map`. It is one of only two tables in core's liquibase schema with no primary key at all (just `concept_name_id` and `concept_name_tag_id`, two indexes and two foreign keys), and the only one of the two that Envers audits, since `ConceptName` is `@Audited` and owns the `tags` collection on it. `filterMappingsWithExistingBase` keeps the mapping because `concept_name_tag_map` does exist, then `executeBackfill` takes `joinColumns` from `getPrimaryKeyColumns` of the base table, gets nothing back, and throws "no shared key columns between base and audit table". `backfillExistingDataIfEnabled` catches that, sets `allSucceeded = false`, and never writes `auditlogweb.backfillExistingData.completed`. So on any stock 2.7/2.8 database the backfill repeats its full `INSERT ... SELECT ... WHERE NOT EXISTS` sweep over all 110 audited tables, `obs` included, on every restart, and config.xml's "Runs only once" stops holding.

Checked against the real core 2.7 metamodel with Envers switched on: `resolveBackfillableTableMappings()` returns 110 core mappings and `concept_name_tag_map -> concept_name_tag_map_AUD` is one of them. I reproduced the throw on H2 with an `@Audited` entity whose join table has no primary key: the mapping is included, `backfillTable` raises exactly that `IllegalStateException`, and the loop finishes with `allSucceeded` false.

I would fall back to the shared data columns as the NOT EXISTS key when the base table has no primary key. Here that is `concept_name_id` and `concept_name_tag_id`, both NOT NULL, so the statement stays exact and the rows actually land. The alternative is to drop no-PK mappings here the way base-less ones are dropped, which at least lets the flag latch, but then concept name tags read as empty at the baseline revision, so I would rather have the data. A case for it in `AuditBackfillDaoIntegrationTest` would be worth adding too, since nothing in the suite calls `backfillTable` today and the insert path is covered only by string comparisons on the generated SQL.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDao.java
```diff
@@ -297,6 +725,77 @@ String buildBackfillInsertSql(TableMapping mapping, List<String> dataColumns, Li
...
+			sql.append(quoteIdentifier(column.name, quote)).append(' ').append(column.typeDdl).append(", ");
+		}
+		sql.append("REV ").append(revColumnType).append(" NOT NULL, REVTYPE ").append(revTypeColumnType);
+		if (!basePkColumns.isEmpty()) {
+			sql.append(", PRIMARY KEY (");
+			for (String pkColumn : basePkColumns) {
+				sql.append(quoteIdentifier(pkColumn, quote)).append(", ");
+			}
+			sql.append("REV)");
+		}
```
dkayiwa, line 895: Worth adding an index on REV here. This is a performance point rather than a correctness one, so it's your call whether it lands in this PR or a follow-up, but it is cheap while this code is fresh and the module's own read path is what pays for it.

`AuditDao.getEntitiesModifiedInRevision` (line 477, behind `AuditServiceImpl.getRelatedEntitiesInRevision`) is the one audit query that filters on the revision alone, with no entity id alongside it:

```java
forRevisionsOfEntity(clazz, false, true)
    .add(AuditEntity.revisionNumber().eq(revisionId))
```

Envers compiles that into a predicate on the audit table itself. Captured from an actual run against an Envers-enabled schema:

```sql
select ... from doc_AUD e cross join revision_entity r
where e.REV = ? and e.REV = r.id order by e.REV asc
```

A table built by this method has exactly one index, `PRIMARY KEY (basePk..., REV)`, and REV is never its leading column, so `e.REV = ?` cannot use it. I checked over JDBC metadata after `createMissingAuditTables()` ran and the entity audit table's only index is `PRIMARY_KEY(id, REV)`. So each of those lookups is a full scan, and the cost tracks table size rather than result size: once the backfill has run, `obs_AUD` holds a row per `obs` row and gets scanned end to end to return one row or none.

The Javadoc just above says this is the same shape Envers' own DDL uses, and that part isn't quite right. Hibernate's DDL for a synthetic audited entity, dumped with MySQL57Dialect since that is what core configures:

```sql
create table doc_AUD (id integer not null, REV integer not null, REVTYPE tinyint, title varchar(255), primary key (id, REV)) engine=InnoDB;
alter table doc_AUD add constraint FK... foreign key (REV) references revision_entity (id);
```

Envers gets a REV index for free out of that foreign key, because InnoDB indexes foreign key columns. Your schema migration pass produces the same thing for the base-less tables, which leaves the resulting schema uneven: `Location_LocationAttribute_AUD` comes out with a REV index and `obs_AUD` does not.

I would keep the no-foreign-keys decision and issue a separate `CREATE INDEX <audit table>_rev ON <audit table> (REV)` after each CREATE TABLE, rather than an inline `INDEX (REV)`, which MySQL accepts but H2 does not, so `AuditBackfillDaoIntegrationTest` keeps running as is. Adding the foreign key instead would also work and would match Envers, but it constrains insert ordering for no gain here.

wikumChamith:
@dkayiwa I updated the PR.

api/src/main/java/org/openmrs/module/auditlogweb/api/dao/AuditBackfillDao.java
```diff
@@ -117,6 +140,523 @@ public List<TableMapping> orderByAuditTableDependencies(List<TableMapping> mappi
...
+					}
+					if (!existingTables.contains(mapping.baseTable.toLowerCase(Locale.ROOT))) {
+						log.debug("No base table {} to sync columns from; skipping {}.", mapping.baseTable,
+						    mapping.auditTable);
+						continue;
+					}
+					try {
+						Set<String> auditColumns = toLowerCaseSet(getColumnNames(metaData, catalog, mapping.auditTable));
+						for (ColumnDefinition column : readColumnDefinitions(metaData, catalog, mapping.baseTable)) {
+							if (auditColumns.contains(column.name.toLowerCase(Locale.ROOT))) {
```
dkayiwa: A column that already exists but carries the wrong type slips through here, and that case needs fixing before this merges.

Core widens an audited column in a 2.8.x maintenance release. ChangeSet `2025-09-30-TRUNK-6245` in `liquibase-update-to-latest-2.8.x.xml` (present from 2.8.1 on, absent in 2.8.0) runs `modifyDataType tableName="concept_reference_range" columnName="criteria" newDataType="TEXT"`, and 2.7.x had created that column as `VARCHAR(255) NOT NULL`. `ConceptReferenceRange` is `@Audited`, so on 2.7.x or 2.8.0 this module clones `concept_reference_range_AUD.criteria` as `varchar(255)`. After the upgrade the base column is `text` and the audit column is not, but `criteria` is present by name, so nothing gets added, `failedTables` comes back empty, and `syncAuditColumnsIfPlatformUpgraded` records the new version, which means the sweep never runs again. The first ConceptReferenceRange saved with a criteria longer than 255 characters then fails its Envers insert and the whole transaction rolls back, so the edit is lost, and nothing at startup had warned. Worth remembering that 255 was widened precisely because it proved too short in practice, so long values are what comes next.

Reproduced against MySQL 8.0.46 on its default sql_mode, driving this DAO through that sequence:

```
audit ddl after clone  = CREATE TABLE `probe_range_AUD` (... `criteria` varchar(255) DEFAULT NULL, `REV` int NOT NULL, ...)
base ddl after upgrade = CREATE TABLE `probe_range` (... `criteria` text, ...)
sync columnsAdded = 0, failed = []
audit ddl after sync   = CREATE TABLE `probe_range_AUD` (... `criteria` varchar(255) DEFAULT NULL, ...)
long save (1000 chars) = FAILED -> MysqlDataTruncation: Data too long for column 'criteria' at row 1
rows base = 1, rows audit = 1   (the 1000-char save rolled back entirely)
```

A server running without STRICT_TRANS_TABLES fails more quietly still: the audit row keeps a truncated `criteria`, so the audit log disagrees with the row it is auditing.

`readColumnDefinitions` already hands back the base column's complete type DDL, so comparing it with the audit column's costs nothing. My preference is to treat a mismatch the way a failed table is treated, logging at error with the table, column and expected type and adding the table to `failedTables`, so the version is not recorded and the next startup tries again. That keeps this method add-only and portable, and it puts an unusable audit column in the startup log rather than in somebody's failed save. Emitting the widening statement (`ALTER TABLE ... MODIFY` on MySQL and MariaDB) is the better end state if you want the module to repair it outright, but that DDL is dialect specific, so I would not fold it in here. Either way, a case in `AuditBackfillDaoIntegrationTest` that creates the base column narrow, widens it, re-syncs and then saves an oversized value would hold this in place; the `subtitle` case next to it only covers the add path.

wikumChamith: @dkayiwa done!

wikumChamith:
@dkayiwa is this ready to get merged :) ?

dkayiwa:
@claude review

api/src/main/java/org/openmrs/module/auditlogweb/api/AuditBackfillService.java
```diff
@@ -37,8 +40,61 @@ public class AuditBackfillService {
...
+	 */
+	public void syncAuditColumnsIfPlatformUpgraded() {
+		if (!EnversUtils.isEnversEnabled()) {
+			log.info("Envers is disabled (hibernate.integration.envers.enabled != true); skipping audit column sync.");
+			return;
+		}
+		AdministrationService administrationService = Context.getAdministrationService();
+		String currentVersion = OpenmrsConstants.OPENMRS_VERSION != null ? OpenmrsConstants.OPENMRS_VERSION : "";
+		String lastSyncedVersion = administrationService.getGlobalProperty(GP_COLUMN_SYNC_VERSION, "");
+		if (currentVersion.equals(lastSyncedVersion)) {
```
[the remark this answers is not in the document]
wikumChamith: Updated this to support module version changes too. As far as I know, this is kind of irrelevant right now though, since none of our module tables support auditing yet :)

wikumChamith:
@claude review

wikumChamith:
> @claude review

@dkayiwa Claude gave a thumbs up. Can I merge this? :)

dkayiwa:
@wikumChamith yes 😊

dkayiwa (approved):

PR #61, opened 2026-08-20
Publish RefApp Docker images with the module and audit-log ESM

wikumChamith:

  - Add Dockerfile.backend/Dockerfile.frontend overlaying the module omod and @openmrs/esm-audit-log-app onto the official O3 RefApp images
  - Add build-docker.yml: tests, then publishes nightly-auditlog (main pushes + nightly cron) and <version>-auditlog (release tags) to Docker Hub
  - Add release.yml using the shared release-backend-module.yml workflow
  - Add production docker-compose.yml (Envers enabled, required passwords) and a dev override that builds locally and mounts the local omod
  - Dockerfile.backend sets hibernate.hbm2ddl.auto=update (and Envers enabled) via OMRS_EXTRA_* env vars, so the Envers schema is created before core's first audited write and kept in step on every boot; overridable at runtime
  - Document setup and usage in the README

## Issue I worked on

see https://openmrs.atlassian.net/browse/AUDIT-

  No? -> execute above command


scripts/hbm2ddl-wrapper.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# First boot runs Hibernate with hbm2ddl.auto=update so the Envers schema
+# exists before core's first audited write — the module's own table creation
+# runs in the activator, which is too late for writes made during core and
+# earlier-module startup. Once a boot reaches healthy, a marker on the data
+# volume switches every later boot to hbm2ddl.auto=none; from then on the
+# module keeps the schema in sync.
```
dkayiwa: This needs fixing before merge, because nothing in the module keeps the audit schema in sync. `AuditlogwebActivator.started()` only calls `backfillExistingDataIfEnabled()`, and `api/src/main/resources/liquibase.xml` creates `audit_security_event`, `read_audit_log` and `read_audit_entity_metadata`, never the Envers `_AUD` tables or `revision_entity`. I grepped `api/src/main` and `omod/src/main` for anything that issues DDL and the only hit is the backfill's `INSERT ... SELECT`. So once the marker is written, the audit schema is frozen at whatever the first boot happened to produce, and nothing repairs it.

Two ways that bites on the stack this PR ships:

1. Boot once with `OMRS_EXTRA_HIBERNATE_INTEGRATION_ENVERS_ENABLED=false`, which `Dockerfile.backend` line 19 explicitly offers as a runtime override. Envers is off, so no audit entities are in the Hibernate metadata and `hbm2ddl.auto=update` creates no `_AUD` tables, but the poller still sees `/health/started` and writes the marker anyway. Turn Envers back on and every later boot runs with `none`, so the first audited save dies on `Table 'openmrs.person_AUD' doesn't exist`. Envers writes its audit row inside the caller's transaction, so that rolls back the save itself: you cannot register a patient.

2. `nightly-auditlog` rebuilds nightly from a base whose core is `openmrs/openmrs-core:2.8.x`, and 114 classes in core 2.8.x carry `@Audited`, `BaseOpenmrsObject` and `BaseOpenmrsData` among them, so effectively the whole domain model is audited. The first core release that adds a mapped property to an audited entity leaves the `_AUD` twin without that column, Envers names it in the audit insert, and you get the same rollback. `<version>-auditlog` has the same exposure the first time a pinned deployment is moved forward over an existing volume.

The module's own admin hint already prescribes the fix. `omod/src/main/resources/messages.properties:15` tells administrators to set `hibernate.integration.envers.enabled=true` **and** `hibernate.hbm2ddl.auto=update`, permanently. So in `Dockerfile.backend`:

    ENV OMRS_EXTRA_HIBERNATE_HBM2DDL_AUTO="update"

and this script, the marker, the background poller, the `CMD` override and the "always reset both volumes together" rule in the README all go away. Whatever risk `update` carries against the Liquibase-managed schema is already taken on the first boot, and since SchemaUpdate only ever adds, re-running it each boot converges instead of drifting. Leaving it on is the thing that makes the audit tables track the mappings.

If you did have a reason to switch it off after the first boot, it needs to be written into this header, because as it stands the header says the module covers it and the README repeats that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+    with:
+      release_version: ${{ inputs.release_version }}
+      development_version: ${{ inputs.development_version }}
+      branch: ${{ inputs.branch }}
+    secrets:
+      MAVEN_REPO_USERNAME: ${{ secrets.MAVEN_REPO_USERNAME }}
+      MAVEN_REPO_API_KEY: ${{ secrets.MAVEN_REPO_API_KEY }}
+      GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+      GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+      BOT_PAT: ${{ secrets.BOT_PAT }}
```
dkayiwa: Does this repo actually have a `BOT_PAT` secret? `maven.yml` reaches for `secrets.OMRS_BOT_GH_TOKEN`, and fhir2 and legacyui both pass `BOT_PAT: ${{ secrets.OMRS_BOT_GH_TOKEN }}` into this same reusable workflow, so the name is not consistent across the org. If it is unset here the expression resolves to an empty string, `release-backend-module.yml` declares `BOT_PAT` optional, and its checkout quietly falls back to `github.token`.

That matters beyond branch protection. A tag pushed with `GITHUB_TOKEN` does not start workflow runs, so the `tags: ['[0-9]+.[0-9]+.[0-9]+']` trigger in `build-docker.yml` would never fire, and the `<version>-auditlog` images this PR exists to publish would silently never get built. I could not read the secret list to settle it myself, hence the question.

docker-compose.dev.yml
```diff
@@ -0,0 +1,29 @@
...
+  backend:
+    image: auditlogweb-backend:local
+    build:
+      context: .
+      dockerfile: Dockerfile.backend
+    volumes:
+      # Update the filename when the pom version changes.
+      - type: bind
+        source: ./omod/target/auditlogweb-1.1.0-SNAPSHOT.omod
+        target: /openmrs/distribution/openmrs_modules/auditlogweb-1.1.0-SNAPSHOT.omod
```
dkayiwa: The `release.yml` this PR adds takes a `development_version` and sets it on `main`, so the first release moves the pom to `1.2.0-SNAPSHOT` and this path stops existing. `create_host_path: false` then turns every dev's `up` into a bind-mount error, and nothing in CI exercises the dev override, so the first person to find out is a contributor. Since `.env.dev` is already in the invocation, one variable keeps it in a single place:

```suggestion
      # MODULE_VERSION comes from .env.dev; keep it in step with the pom version.
      - type: bind
        source: ./omod/target/auditlogweb-${MODULE_VERSION}.omod
        target: /openmrs/distribution/openmrs_modules/auditlogweb-${MODULE_VERSION}.omod
```

with `MODULE_VERSION=1.1.0-SNAPSHOT` added to `.env.dev`. Not a blocker, but the line as written breaks at the next release rather than eventually.

spa-assemble-config.json
```diff
@@ -0,0 +1,51 @@
+{
+  "frontendModules": {
+    "@openmrs/esm-active-visits-app": "next",
+    "@openmrs/esm-appointments-app": "next",
+    "@openmrs/esm-audit-log-app": "next",
```
dkayiwa: Did you consider deriving this list at build time rather than vendoring it? I diffed the file against `openmrs-distro-referenceapplication/frontend/spa-assemble-config.json` and it is byte for byte that file plus this one line (the only other change is the dropped `excludedFrontendModules`, which `assemble` ignores anyway, since the key it actually reads is `frontendModuleExcludes`). Keeping a copy means it goes stale silently: when the distro adds an app, `nightly-auditlog` simply will not have it, nothing fails, and the image quietly stops being "the RefApp plus the audit log app", which is the only reason it exists.

One way out, given you are already in a multi-stage build: assemble just `@openmrs/esm-audit-log-app` into a scratch target, pull the base image's own `importmap.json` and `routes.registry.json` into the node stage with `COPY --from=openmrs/openmrs-reference-application-3-frontend:nightly`, merge the single new entry into each, then copy the merged files and the new module directory over the base. The other 46 apps then come from whatever base you layer on, by construction. That is more Dockerfile than this file is, so your call, but the drift is real either way.

wikumChamith:
@dkayiwa I updated the PR.

docker-compose.yml
```diff
@@ -0,0 +1,93 @@
...
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost/"]
+      timeout: 5s
+    depends_on:
+      - backend
+    logging: *logging
+
+  backend:
+    # Built by .github/workflows/build-docker.yml from Dockerfile.backend
+    # (RefApp backend + auditlogweb module + first-boot Envers schema wrapper).
```
dkayiwa: The wrapper this names went away in the same push that added the file, so anyone who goes looking for it comes up empty. The image now just sets the two Envers properties.

```suggestion
    # (RefApp backend + auditlogweb module + Envers enabled).
```

README.md
```diff
@@ -52,3 +52,56 @@ If uploads are not allowed from the web (changable via a runtime property), you
...
+and published to Docker Hub by the `build-docker.yml` workflow — the same setup
+as openmrs-module-chartsearchai. Pushes to `main` and a nightly schedule publish
+`openmrs/openmrs-reference-application-3-{backend,frontend}:nightly-auditlog`;
+release tags publish `<version>-auditlog` (e.g. `1.1.0-auditlog`) for pinned
+deployments. The dev compose override builds the images locally under separate
+tags, so it works before any image is published.
+
+Prerequisites: Docker with the compose plugin.
+
+**Production:**
```
dkayiwa: This heading is worth reconsidering, because everything under it except the module version rides a moving tag. Both Dockerfiles build `FROM openmrs/openmrs-reference-application-3-{backend,frontend}:nightly`, the frontend resolves every ESM at `next`, and the gateway in `docker-compose.yml` stays on `${GATEWAY_TAG:-nightly}`. Someone who follows this section is therefore running unreleased O3 with pre-release ESMs, and the `<version>-auditlog` images described above as "for pinned deployments" pin the module and nothing else: those two images are immutable once built, but the gateway tag underneath them is not, so a later `docker compose pull` can move it.

Cheapest fix is to say that here and retitle the section, since what this stack really offers is an evaluation/demo deployment. If you want the release-tag images to promise more, an `ARG BASE_TAG=nightly` in both Dockerfiles with the current RefApp release tag passed on tag builds, plus a matching `GATEWAY_TAG` default, gets there, though that is more than this PR needs. I would not hold the PR for either.

.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,100 @@
...
+# and <version>-auditlog on the release tags created by release.yml (the
+# parent pom tags releases as the plain version, e.g. 1.1.0).
+on:
+  push:
+    branches: [main]
+    tags: ['[0-9]+.[0-9]+.[0-9]+']
+  schedule:
+    # Nightly, so the "next" ESMs and nightly base images don't go stale.
+    - cron: '45 2 * * *'
+  workflow_dispatch:
```
dkayiwa, line 21: There is no `pull_request` trigger, so neither Dockerfile gets built until a push to `main`, and that same run publishes. The first time the frontend's `ADD` plus `node -e` plus `assemble` chain, or the backend's in-image Maven build, actually runs is after review has closed, straight onto the tag the README tells people to `docker compose up`. `maven.yml` runs on PRs but only touches Maven, so nothing in CI reads these two files before then. The consequence is not just a red run after the fact: frontend and backend publish from two independent jobs, so if one fails once the other has pushed, `nightly-auditlog` is left as a mismatched pair until someone notices and re-runs it.

Adding `pull_request: branches: [main]` here, with the two publish steps taking `push: ${{ github.event_name != 'pull_request' }}` and a single platform on PRs, catches the broken-Dockerfile case before it can do that (it will not help with a transient registry failure, but that is the one you can re-run). Nothing gets pushed on a PR, so the Docker Hub login can be skipped there too and it keeps working for forks. Worth doing as one change rather than adding the trigger by itself, since the trigger alone would have PRs publishing.

Not a blocker, but this PR is entirely build files and this would be the only coverage they have.

.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,100 @@
...
+
+permissions:
+  contents: read
+
+env:
+  IMAGE: openmrs/openmrs-reference-application-3
+  DOCKER_IMAGE_TAG: ${{ github.ref_type == 'tag' && format('{0}-auditlog', github.ref_name) || 'nightly-auditlog' }}
```
dkayiwa, line 28: `openmrs/openmrs-reference-application-3-backend:nightly-auditlog` already exists on Docker Hub. It was pushed on 18 Aug alongside `3.7.x-auditlog` and `audit`, all three within four seconds, which is the three-tag shape `build-backend.yml` in openmrs-distro-referenceapplication produces from its `docker_image_tag` / `docker_image_tag_version` / `docker_image_tag_env` inputs. I pulled the image config for that tag and its layer history is the distro Dockerfile's, with no omod copy and no `OMRS_EXTRA_*` env, so it did not come from `Dockerfile.backend`. The frontend repo has no `-auditlog` tag at all.

So something on the distro side is already publishing this name. Is this workflow meant to take it over? If so, the other two are worth deleting or re-pointing: they will sit frozen at the 18 Aug build while `nightly-auditlog` moves nightly, and anyone deploying `audit` or `3.7.x-auditlog` will reasonably assume the three match. If the distro build is meant to keep producing it as well, then whichever pipeline ran last owns the tag and a `docker compose pull` becomes a coin flip, and this would be better off on a tag this repo owns.

wikumChamith:
@dkayiwa  I updated PR.

Dockerfile.backend
```diff
@@ -0,0 +1,25 @@
...
+
+# The configuration the module's admin hint prescribes (messages.properties):
+# Envers on, and hbm2ddl.auto=update permanently so the Envers schema is
+# created before core's first audited write and re-checked against the audited
+# mappings on each boot (SchemaUpdate only ever adds). Note SchemaUpdate logs
+# DDL failures and continues, so missing _AUD tables surface in the backend
+# log, not as a startup failure. Both overridable at runtime; set HBM2DDL to
+# none when running multiple backend replicas (concurrent DDL races).
+ENV OMRS_EXTRA_HIBERNATE_INTEGRATION_ENVERS_ENABLED="true"
+ENV OMRS_EXTRA_HIBERNATE_HBM2DDL_AUTO="update"
```
dkayiwa, line 25: I booted this image against the compose file in this PR to see what `update` actually does to the audit schema, and it is worth knowing before the tag goes out. SchemaUpdate builds the `_AUD` tables from the entity mappings while the base tables come from Liquibase, so the two disagree. On a first boot the module's own sweep flags 95 of the 113 audit tables as diverging and logs an ERROR for each, and since `syncAuditColumnsIfVersionsChanged` only records its fingerprint after a clean sweep, `auditlogweb.auditColumnSync.versionFingerprint` is still empty afterwards, so the whole sweep repeats on every boot.

Most of the divergences are cosmetic (`uuid` lands as VARCHAR(38) against a CHAR(38) base, several text columns land wider). 24 columns land narrower than the column they mirror, though: `notification_template_AUD.template` is varchar(255) against a `text` base, and `order_set_AUD.description` and `serialized_object_AUD.retire_reason` are varchar(255) against varchar(1000). The MariaDB in `docker-compose.yml` runs with STRICT_TRANS_TABLES, so a value that fits the base column but not its audit twin makes the Envers insert fail at commit, and because Envers writes inside the caller's transaction that rolls the user's save back with it. Nothing in the demo data is long enough to hit it today, so it is latent rather than broken.

I am not asking you to drop the property. `AuditBackfillDao` derives audit columns from the base table's JDBC metadata, so tables it creates match exactly, but it runs from the activator and that is too late: with HBM2DDL set to `none` I get 109 `Table 'openmrs.revision_entity' doesn't exist` errors and 22 failures out of `Context.checkCoreDataset` before the module ever starts. Core 2.8.7 in the `nightly` base has no audit-table initializer of its own (2.9 adds `EnversAuditTableInitializer`, schema-filtered to the audit tables, which is what will eventually make this go away).

What would help is the README's "Verify the audit schema" note, which tells operators to check the backend log for `SchemaUpdate` errors if tables are missing. As shipped, that log always carries 95 audit-table ERRORs, so it is worth saying there that they are expected and what they mean. Widening the narrow columns looks like an AUDIT-62 follow-up rather than anything this PR can fix.

Dockerfile.backend
```diff
@@ -0,0 +1,25 @@
...
+
+COPY --from=builder /build/omod/target/auditlogweb-*.omod /openmrs/distribution/openmrs_modules/
+
+# The configuration the module's admin hint prescribes (messages.properties):
+# Envers on, and hbm2ddl.auto=update permanently so the Envers schema is
+# created before core's first audited write and re-checked against the audited
+# mappings on each boot (SchemaUpdate only ever adds). Note SchemaUpdate logs
+# DDL failures and continues, so missing _AUD tables surface in the backend
+# log, not as a startup failure. Both overridable at runtime; set HBM2DDL to
+# none when running multiple backend replicas (concurrent DDL races).
```
dkayiwa, line 23: On a fresh database this is the configuration that fails. I ran it: with `OMRS_EXTRA_HIBERNATE_HBM2DDL_AUTO=none` and Envers on, nothing creates `revision_entity` before core's own startup writes, so `Context.checkCoreDataset` goes down with "Error while setting core privileges" and "Error while setting core global properties" (109 `Table 'openmrs.revision_entity' doesn't exist` errors in that one boot), and the activator never gets far enough to create the audit tables either. Setting it to `none` is only safe once the audit schema is already there, which is worth spelling out, because an env var set on a deployment reaches every replica including the one that boots first.

```suggestion
# log, not as a startup failure. Both overridable at runtime; once the audit
# schema exists, extra backend replicas can set HBM2DDL to none to avoid
# concurrent DDL races, but never set it on a fresh database.
```

.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,136 @@
...
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+
+      - name: Retag candidates as the published pair
+        run: |
+          docker buildx imagetools create -t "$IMAGE-frontend:$DOCKER_IMAGE_TAG" "$IMAGE-frontend:$CANDIDATE_TAG"
+          docker buildx imagetools create -t "$IMAGE-backend:$DOCKER_IMAGE_TAG" "$IMAGE-backend:$CANDIDATE_TAG"
```
ibacher, line 136:
```suggestion
          docker buildx imagetools create -t "$IMAGE-backend:$DOCKER_IMAGE_TAG" "$IMAGE-backend:$CANDIDATE_TAG"

```

.github/workflows/release.yml
ibacher: `OMRS_BOT_GH_TOKEN` is the name of the org secret so `BOT_PAT` will always be an empty string. The org secret is not a `GITHUB_TOKEN` and we use it precisely to be able to trigger downstream builds.

.github/workflows/build-docker.yml
wikumChamith: @ibacher, is there a difference between these two? :)

Dockerfile.frontend
```diff
@@ -0,0 +1,28 @@
...
+# Use the official O3 frontend image as the base — it has the correct
+# index.html (with SPA bootstrap), nginx config, and env var substitution.
+# Note the app shell comes from the base image while modules resolve at build
+# time, so the two can drift between base-image and module releases.
+FROM openmrs/openmrs-reference-application-3-frontend:nightly
+# Drop the base image's app bundles first (the shell, fonts, and config files
+# live at the html root and survive): the overlay below brings its own set,
+# and stale copies of distro-removed apps would otherwise linger on disk.
+RUN rm -rf /usr/share/nginx/html/openmrs-esm-*
+COPY --from=build-spa /app/dist /usr/share/nginx/html
```
dkayiwa: This needs fixing before merge. As built, the image never loads the audit-log app in a browser: nginx serves the base image's stale compressed importmap, which doesn't name the ESM.

The base's nginx prefers a precompressed sibling over the plain file. Its `/etc/nginx/nginx.conf` gives `location ~* \.json$` a `try_files $uri$enc_suffix $uri$enc_fallback $uri =404`, where `$enc_suffix` is mapped from `Accept-Encoding` and `.br` is tried first. `openmrs assemble` writes no `.gz`/`.br` at all (I checked the build stage output: zero of either under `/app/dist`), so the `COPY` on the next line replaces `importmap.json` and `routes.registry.json` while the base's siblings stay beside them. The `rm -rf` here doesn't reach those, since they sit at the html root rather than inside the module directories.

Built the image and ran it to be sure. `curl -H 'Accept-Encoding: gzip, deflate, br' /importmap.json` comes back `Content-Encoding: br`, with a body whose sha256 matches the base image's `importmap.json.br` exactly: 47 entries, no `@openmrs/esm-audit-log-app`. `routes.registry.json` does the same, so the app gets no route or extension either. Drop the header and the fresh 48-entry file is served, which is why curl checks pass and only a real browser fails. The ESM's own bundle is reachable at 200, nothing ever imports it.

openmrs-module-chartsearchai, which the README cites as the model for this setup, ran into exactly this and its `Dockerfile.frontend` carries the write-up. Worth taking their treatment whole: delete the base's `.gz`/`.br` siblings before the `COPY`, run their `precompress-spa.mjs` over `/app/dist` in the `$BUILDPLATFORM` stage (compressing in the final stage would do it under qemu for `linux/arm64` on every build), and keep their build-time grep of `importmap.json.gz` for the ESM name, which is the check this workflow's green Docker build doesn't make.

Deleting just `importmap.json.{br,gz}` and `routes.registry.json.{br,gz}` on this line clears the blocker too (I tested that), but it leaves everything serving uncompressed: `openmrs-esm-patient-chart-app.js` goes out at 33,684 bytes here against 8,883 brotli from the stock base image.

.github/workflows/build-docker.yml
dkayiwa: The file has no newline at the end, it stops right after the closing quote on line 136, so the suggestion adds one. The command text is byte for byte what's already there.

Dockerfile.frontend
wikumChamith: Updated

Dockerfile.frontend
```diff
@@ -0,0 +1,60 @@
...
+# config files live at the html root and survive; stale copies of
+# distro-removed apps would otherwise linger on disk) and ALL of its
+# precompressed siblings. The siblings matter most: nginx prefers <file>.br /
+# <file>.gz over the plain file, so a stale sibling beside a replaced
+# importmap.json or routes.registry.json would hand every browser the base's
+# module list — which never names the audit-log app — while a plain curl
+# (no Accept-Encoding) still reads the fresh file and looks healthy.
+RUN rm -rf /usr/share/nginx/html/openmrs-esm-* && \
+    find /usr/share/nginx/html \( -name '*.gz' -o -name '*.br' \) -delete
+COPY --from=build-spa /app/dist /usr/share/nginx/html
```
dkayiwa: This needs fixing before merge. Deleting every precompressed sibling in the tree also removes the ones for files the copy never replaces, and nothing regenerates them, so the published image serves the app shell raw: 3.6 MB per page load where the stock base sends 148 KB.

`openmrs assemble` writes only `importmap.json`, `routes.registry.json` and the 48 module directories, so the shell, its chunks, the fonts and `config-core_demo.json` are all base-only, and `precompress-spa.mjs` runs over `/app/dist`, which never holds them.

I built 26a6478 and served it. The two files `index.html` pulls on every request come back with no `Content-Encoding`: `openmrs.c7aee46767816c4f.css` at 3,454,392 bytes and `openmrs.94d7f8d538362997.js` at 142,692, against 114,814 and 33,147 brotli from the stock `nightly` base. There is no `gzip on` anywhere in the image's nginx.conf, so nothing compresses them on the fly. Across the html root that is 67 negotiable .js/.css/.json/.svg files holding 7,597,457 uncompressed bytes, where the base has 726 such files holding 415,226. chartsearchai carries the same gap (its `spa-build-config.json` has no app shell either), so it isn't a precedent for this part.

Driving the deletion off the assembly's own file list fixes it without weakening the guard below. I built and ran this version: `importmap.json` still comes back `Content-Encoding: br` with 48 entries including the audit-log app, the shell CSS is back to 114,814 br, and the uncompressed set drops to 2 files holding 1,462 bytes.

```suggestion
# Before the copy: drop the base image's app bundles (the shell, fonts, and
# config files live at the html root and survive; stale copies of
# distro-removed apps would otherwise linger on disk), then drop the
# precompressed siblings of just the files the copy replaces. nginx prefers
# <file>.br / <file>.gz over the plain file, so a stale sibling beside a
# replaced importmap.json or routes.registry.json would hand every browser the
# base's module list, which never names the audit-log app, while a plain curl
# (no Accept-Encoding) still reads the fresh file and looks healthy.
# Driving that off the assembly's own file list, rather than deleting every
# sibling in the tree, keeps the base-only files compressed: nothing here
# rebuilds them and this stage has no brotli compressor.
RUN rm -rf /usr/share/nginx/html/openmrs-esm-*
RUN --mount=type=bind,from=build-spa,source=/app/dist,target=/dist \
    cd /dist && find . -type f ! -name '*.gz' ! -name '*.br' \
      -exec rm -f "/usr/share/nginx/html/{}.gz" "/usr/share/nginx/html/{}.br" \;
COPY --from=build-spa /app/dist /usr/share/nginx/html
```

precompress-spa.mjs
```diff
@@ -0,0 +1,124 @@
...
+    utimesSync(sibling, atime, mtime);
+    written[extension]++;
+  }
+}
+
+const summary = encoders.map((e) => `${written[e.extension]} ${e.extension}`).join(', ');
+console.log(
+  `precompressed ${considered} file(s) under ${dist}: ${summary}` +
+    ` in ${Math.round((Date.now() - startedAt) / 1000)}s`,
+);
```
dkayiwa: No newline at the end of this file, the same thing ibacher caught on `build-docker.yml`. chartsearchai's copy has one, so it was lost on the way over.

```suggestion
);

```

.github/workflows/build-docker.yml
dkayiwa: Re-checked this today. The three tags are still exactly as they were on 18 Aug, and nothing in openmrs-distro-referenceapplication publishes them on a schedule: its `build-backend.yml` is `workflow_dispatch`/`workflow_call` only, with `docker_image_tag`/`docker_image_tag_version`/`docker_image_tag_env` defaulting to nightly/3.7.x/dev3, and grepping its whole `.github/` tree for "audit" returns nothing. So those three came from someone typing the inputs by hand once, and the coin-flip worry above doesn't apply, nothing on the distro side will overwrite what this workflow pushes.

What is left is smaller. Once this merges, `nightly-auditlog` on the backend starts moving nightly while `3.7.x-auditlog` and `audit` sit frozen at that 18 Aug build beside it, so those two are worth deleting or re-pointing.

Related, since it lands on the same shared repos: `promote` retags the candidates but never removes them, so `nightly-auditlog-candidate` stays on `openmrs/openmrs-reference-application-3-{frontend,backend}` permanently and each release adds a `<version>-auditlog-candidate` alongside it. A Hub API delete at the end of the promote job, or pushing candidates to a repo this project owns, keeps that out of the official namespace.

wikumChamith:
The build failure we are seeing now is not related to the PR. It looks like an upstream npm ecosystem breakage

README.md
```diff
@@ -52,3 +52,71 @@ If uploads are not allowed from the web (changable via a runtime property), you
...
+The dev compose override builds the images locally under separate tags, so it
+works before any image is published.
+
+Prerequisites: Docker with the compose plugin.
+
+**Run the published images (demo/evaluation):**
+```
+cp .env.example .env       # then edit .env and set OMRS_DB_PASSWORD and MYSQL_ROOT_PASSWORD
+docker compose up -d
+docker compose logs -f backend   # first boot takes a few minutes
```
dkayiwa: `docker-compose.yml` in this PR puts the first boot at "liquibase install + demo data can exceed 30 min" and gives the backend a one hour `start_period` to cover it, so "a few minutes" here is the number that makes someone think it has hung and start over. The gateway answers 502 for that whole stretch, which makes it look worse than it is.

```suggestion
docker compose logs -f backend   # first boot can take 30+ minutes
```

README.md
```diff
@@ -52,3 +52,71 @@ If uploads are not allowed from the web (changable via a runtime property), you
...
+new build (`up -d` alone won't restart an unchanged container):
+```
+mvn clean package
+docker compose --env-file .env.dev -f docker-compose.yml -f docker-compose.dev.yml up -d --force-recreate backend
+```
+Pass `--build` when the Dockerfiles change — compose only builds automatically
+when the image doesn't exist yet. (The frontend's module list is fetched from
+the RefApp distro at build time, with the audit-log app added.) A compose
+error about the omod bind-mount path means `MODULE_VERSION` in `.env.dev` has
+drifted from the pom version.
```
dkayiwa: Someone who follows the Development section above has no `.env`, only `.env.dev`, so every plain `docker compose ...` after that point stops before it reaches Docker: it cannot interpolate `${OMRS_DB_PASSWORD:?...}` in `docker-compose.yml` and prints `required variable OMRS_DB_PASSWORD is missing a value: set OMRS_DB_PASSWORD in .env`. That takes out both blocks below, the schema check and both `down` commands.

I ran it with compose v2.13.0 against the files in this PR, in a directory holding only `.env.dev`: bare `docker compose exec db ...` and `docker compose down` both fail that way, and `--env-file .env.dev` on its own clears them (the two `-f` flags are not needed for exec or down). It is only the docs, but it is the path a new contributor takes first.

```suggestion
drifted from the pom version.

Every compose command in this flow needs `--env-file .env.dev` as well, the
Verify and Stop / reset blocks below included; without it compose cannot
interpolate the passwords the base file requires and exits before reaching
Docker.
```

wikumChamith:
@dkayiwa is this good to go?

.github/workflows/build-docker.yml
dkayiwa: Last thing still open here. As written, the first push to main leaves `nightly-auditlog-candidate` on the official frontend and backend repos permanently, and every release adds a `<version>-auditlog-candidate` beside it. A Hub delete at the end of the promote job, or candidates pushed to a repo this project owns, keeps them out of that namespace.

Happy for it to be a follow-up if you'd rather, I would just like it settled before the tags actually appear rather than after. The other half, deleting or re-pointing `3.7.x-auditlog` and `audit` so they don't sit frozen at the 18 Aug build once `nightly-auditlog` starts moving, is Hub housekeeping rather than anything in the diff.

wikumChamith: All three tags came from a one-off push to the auditlog branch of openmrs-distro-referenceapplication. I deleted that branch, but I don't have permission to delete the DockerHub images. @dkayiwa. Can you delete the stale audit and 3.7.x-auditlog tags from openmrs/openmrs-reference-application-3-backend on Docker Hub ?

dkayiwa (approved):

